Issue #6, opened 2018-07-22
Review of the chapters 1, 2, and 3


pavel-kirienko:
Here comes an early draft for an early review. This PDF is compiled directly from the current master: [UAVCAN_Specification.pdf](https://github.com/UAVCAN/specification/files/2217431/UAVCAN_Specification.pdf)

I would like @kjetilkjeka and @thirtytwobits to have a critical look. I am going to review everything once again when the content is finished to fix the remaining styling and wording issues.

The DSDL chapter is rather different from what is published on the website right now: TAO is gone, and the serialization section is moved up.

pavel-kirienko:
Updated [UAVCAN_Specification.pdf](https://github.com/UAVCAN/specification/files/2220864/UAVCAN_Specification.pdf)

scottdarch:
[UAVCAN_Specification_thirtytwo.pdf](https://github.com/UAVCAN/specification/files/2221806/UAVCAN_Specification_thirtytwo.pdf)

Comments are in the document. Let me know if this review process works.

I stopped at chapter 3 because I didn't realize that the datatype identifiers for top-level types were optional and this has me really confused. How do you envision such a feature is used? If a type id is determined at runtime then which node determines it and how do the other nodes know what typeid was chosen? I feel like I'm missing something big here because I can't envision how this could possibly work.

pavel-kirienko:
@scottdarch let's move to https://github.com/UAVCAN/specification/pull/7

> Comments are in the document. Let me know if this review process works.

It does.

kjetilkjeka:
I'm sorry this took so long @pavel-kirienko I've been away the last week and will be away until Wednesday. I managed to write some comments from the last draft you linked. I will be able to review/discuss more on Thursday. 

uavcan review:
- 1.2 is makes me belive that datatypes can be changed without incrementing minor version. It should be changed to make it clear that datatypes are immutable. Section 2. explains this well.
- There are DSDL definitions that break the nice layering. The most important is the NodeStatus message. The protocol is in fact dependent on this definition and it can not be updated without breaking the protocol. It should be mentioned that the NodeStatus message is special somwhere (towards the end perhaps) of section 2.
- What is the status of changing up namespaces to reflect functions/domain (aero, ground, etc)? This is not done in your draft. We should probably create an issue or similar to discuss this, right?
 - 3.6.2.2 are unions fixed or variable sized data structures. I think dynamic? In that case, there's an error in the text.
[...]

Issue #4, opened 2018-07-18
Proposed: New High-Level Definition of UAVCAN


thirtytwobits:
![image](https://user-images.githubusercontent.com/19180861/42913775-a5383324-8aab-11e8-8160-a40b36edee34.png)

Changes:

1. Core data types are distinct and required.
2. Standard data types (e.g. equipment) are distinct and optional.
3. Custom data types are first class members of the specification.
4. The interface between the serialization of data types and the transport layer is well-defined.

Item 4 was controversial from the google group start of this proposal but I still think there is value to allowing for a different data serialization layer to sit on top of some UAVCAN implementations (e.g. using libuavcan but not DSDL). I'd like to at least propose it for the record here in our specification repository.

pavel-kirienko:
"Standard functions" in the context of the specification refers to the high-level features of the protocol. As such, this block should rest on top of "Core data types". Unless I'm misreading your proposal.

I am considering to decimate the set of standard data types from the `uavcan.equipment` namespace; right now it seems to be poorly organized and some of the definitions might be overly application-specific. I think we should postpone this until the transport layer is settled and agreed upon.

thirtytwobits:
Updated the diagram to put the Standard Functions above the Core data types

pavel-kirienko:
I suggest closing this as the relevant suggestions have already been incorporated into the current specification draft.

thirtytwobits:
I still don't like your diagram Pavel ;-) Can we keep this open until we finish debating the proposal I put in my latest specification review around three classes for data types; required, optional, and standard?

PR #7, opened 2018-07-25
Review #6

pavel-kirienko:
The resulting PDF with the proposed changes from #6 attached: [UAVCAN_Specification.pdf](https://github.com/UAVCAN/specification/files/2228743/UAVCAN_Specification.pdf)

The following changes have not been implemented:

> Time is meaningless without releases. We should specify that the type will be removed with the next minor release after two years of deprecation. If a type is then deprecated but no further releases occur for that major version it is not removed. We should furthermore reserve the right to remove all deprecated types from a previous major release in the next regardless of time elapsed.

The objective was rather the opposite. Equipment vendors do not care about versions, they care about the longevity of the data types they use. There must be some sort of a time-bound guarantee that the data types that were used to build a product today will remain valid and usable for a predefined period of time. In this setting, the vendor doesn't care how many protocol revisions will happen in that time frame.

Furthermore: although this is debatable, it may not be necessary to tie data type modifications with releases - with the new versioning proposal in place we can manage standard data types independently from the rest of the specification as long as the specification itself is not affected by that. For example, none of the data types defined in the `uavcan.equipment` namespace are mentioned in the specification (although there are a few examples that merely refer to the fact of their existence), so they can be modified within the same specification release. On the other hand, fundamental data types like `uavcan.protocol.NodeStatus` are extremely coupled with the fundamentals of the protocol, so they are unlikely to be changeable unless the specification is updated accordingly.

What do you think?

> except for the !response bit, correct?

The response bit implements the distinction between requests and responses, it's a low-level concept, it is not even mentioned in that part of the specification because it is concerned with a very high-level overview. The response bit is going to be introduced in the chapter "CAN bus transport layer".

> https://github.com/UAVCAN/specification/issues/4

The diagram has been updated; not to the exact version you provided, but a slightly modified one. The only meaningful difference is that I avoided the separation of Standard data types from Core data types. Considering the fact that there is only one mandatory data type (`uavcan.protocol.NodeStatus`), such separation does not offer much value. Do you agree?

> I stopped at chapter 3 because I didn't realize that the datatype identifiers for top-level types were optional and this has me really confused. How do you envision such a feature is used? If a type id is determined at runtime then which node determines it and how do the other nodes know what typeid was chosen? I feel like I'm missing something big here because I can't envision how this could possibly work.

You're seemingly talking about *default data type IDs*. They are provided for all standard data types; as stated in the new specification, default DTIDs are not to be changed for the standard data types as that jeopardizes compatibility.

Vendor-specific data types, however, do not have a centralized way of agreeing upon their data type ID values, because there aren't (and seemingly can't be, although this question is probably worth exploring) a global repository of vendor-specific types. Therefore, items of equipment from different vendors are prone to re-using the same data type identifiers for different vendor-specific messages. This is why the specification requires vendors to provide for a way of reconfiguring the data type ID mapping for all vendor-specific data types. Here's a relevant excerpt from the section 3.7:

```tex
All UAVCAN nodes must use the same data type ID mapping for the standard data types,
as defined by the default data type ID values provided for each of the standard data types.
Since the standard data type ID mapping is immutable, all standard-compliant nodes can always
use standard data types conflict-free.

Vendor-specific data types, however, do not enjoy the lack of conflict guarantee,
because by virtue of being vendor-specific, such data types cannot use a global
fixed agreed upon mapping like the standard data types do.
As such, whenever vendor-specific data types are used, there is always a risk that
different nodes may map different data types to the same data type ID.

It is the responsibility of the system integrator to ensure that if vendor-specific data types are
used, the data type ID mappings are configured on all nodes identically.
Vendors of UAVCAN equipment must provide the integrator with a way to change the data type ID
of any vendor-specific data type leveraged by the node\footnote{Nodes that fail to provide a way of
altering the data type ID mapping for vendor-specific data types cannot be considered standard-compliant}.
```


pavel-kirienko:
I found that it is necessary to have this merged so that I could continue with my work on master. So I am merging this now. We can get back to the review later.

PR #12, opened 2018-08-12
[WIP] Transport agnostic spec

pavel-kirienko:
This is per the suggestion made by @kjetilkjeka: 

> I think it would be better to avoid introducing can before neccessary. I would even refrain from calling the section "CAN bus transport layer", instead i would call it "The transport layer". Then first every concept of the transport frame could be defined (this is basically 4.1+4.2+4.4 just use the word transport frame instead of CAN Frame). Then the CAN transport layer could be it's own section (e.g 5) or subsection(s) (e.g 4.4). If this route is chosen, a definition similar to the following should exist - "Transport Frame - a frame of the underlying transport protocol (e.g CAN2.0B or CAN-FD)."

The specification (relative to https://github.com/UAVCAN/specification/pull/11) has been reorganized to separate all CAN-bus-specific particulars into a well-contained section.

Next, I am going to move the CAN-specific hardware design recommentations (connector types, bit rates, etc) into the CAN-specific section, and add a few notes on the [inner priority inversion problem](https://github.com/UAVCAN/uavcan.rs/issues/7#issuecomment-334442123).

The latest build attached: [UAVCAN_Specification.pdf](https://github.com/UAVCAN/specification/files/2281238/UAVCAN_Specification.pdf)

There was a lot of drastic changes to the document, so at this point it's probably hard to read and is not very well organized. I'm going to proof-read everything once again when the changes are merged into master.


specification/UAVCAN_Specification.tex
```diff
@@ -35,7 +38,7 @@ \section*{Overview}
     \item Efficient exchange of large data structures with automatic decomposition and reassembly.
     \item Lightweight, deterministic, easy to implement, and easy to validate.
     \item Suitable for deeply embedded, resource constrained, hard real-time systems.
-    \item Doubly- or triply- redundant CAN bus support.
+    \item Supports doubly- and triply-redundant transports.
```
kjetilkjeka: Is this wording ambigous? I think double redundant may mean 3 can busses (two may fail) and triply redundant may mean 4 (three may fail)?

pavel-kirienko: Rephrased

Issue #16, opened 2018-08-21
Review chapters 1, 2, 3, 4, 6


pavel-kirienko:
The spec draft is ready except for the chapter 5 (Application layer). The work on the application layer is blocked by the lack of the new standard data type set, so that seems to be the next thing to do.

The document has been restructured rather considerably compared to the v0 spec published on the website. The new structure provides dedicated chapters for the transport layer, the application layer, and the physical layer. Also, per Kjetil's suggestion, CAN-specific information has been extracted into two well-isolated sections (one for the transport layer and the other for the physical layer), allowing for future support of other transports.

The CAN FD physical layer compatibility recommendations are missing. They are going to remain this way until we've ran some tests with CAN FD in the lab here; nevertheless, relevant discussions and suggestions are highly welcome.

The current draft could probably use some refinement in the following areas (help needed):

- **The specification of the DSDL grammar (sections 3.2, 3.3, 3.4) is rather haphazard.** I believe it could greatly benefit from a complete rewrite from scratch. The main reason it ended up in this state is that the early draft was updated and amended in various ways multiple times until the current edition was settled upon. Now, having the design solidified, is a good opportunity to rework it into a proper specification. It is not a high priority task but is something I could use help with.
- **Section 4.3 (transport layer - transfer reception) could be incomplete or hard to understand.** Transfer reconstruction is probably the hardest part of the protocol to implement right, so it is important to ensure that this part is comprehensible and is unambiguous.

Here is the latest draft (built from the current master): [UAVCAN_Specification.pdf](https://github.com/UAVCAN/specification/files/2307197/UAVCAN_Specification.pdf)

I think as we are slowly approaching completion of the document I should stop cowboying and begin using a more disciplined approach to editing with feature branches, especially if you're going to edit the document alongside.

kjetilkjeka:
This is a pretty big issue and I'm probably going to "proofread" and bring up concerns a bit spordically. Since this is a big issue we should probably let this branch out into different issue. On top of my head right now i have:
- We should discuss "3.6 Data type compatibility and versioning" some more in #9 
- Improve the readability of the CAN ID structure #17

thirtytwobits:
I have started an in-depth review of this version. I'll post a PDF with my notes when I complete it. This could take awhile.

pavel-kirienko:
[...]

PR #21, opened 2018-09-07
#16 changes for chapter 1

thirtytwobits:
Mostly English changes but a few corrections.

### DIFF
--- a/specification/introduction/introduction.tex
+++ b/specification/introduction/introduction.tex
@@ -8,73 +8,66 @@ The actual specification is contained in the following chapters.
 UAVCAN is designed to adhere to the following set of basic principles.
 
 \begin{description}
-    \item[Democratic network] - There will be no master node.
+    \item[Democratic network] --- There will be no master node.
     All nodes in the network will have the same communication rights; there must be no single point of failure.
 
-    \item[Nodes can exchange long payloads] - Nodes must be provided with a simple way of exchanging large
+    \item[Nodes can exchange long payloads] --- Nodes must be provided with a simple way of exchanging large
     data structures that cannot fit into a single transport frame (such as GNSS solutions, 3D vectors, etc.).
     UAVCAN should perform automatic transfer decomposition and reassembly at the protocol level,
     hiding the related complexity from the application.
 
-    \item[Support for redundant interfaces and redundant nodes] - UAVCAN must be suitable for use in
+    \item[Support for redundant interfaces and redundant nodes] --- UAVCAN must be suitable for use in
     applications that require modular redundancy.
 
-    \item[High throughput, low latency communication] - UAVCAN must be suitable for use in
+    \item[High throughput, low latency communication] --- UAVCAN must be suitable for use in
     hard real-time systems.
 
-    \item[Simple logic, low computational requirements] - UAVCAN targets a wide variety of embedded systems,
+    \item[Simple logic, low computational requirements] --- UAVCAN targets a wide variety of embedded systems,
     from high-performance embedded on-board computers for intensive data processing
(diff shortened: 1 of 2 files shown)


specification/introduction/introduction.tex
```diff
@@ -8,73 +8,69 @@ \section{Design principles}
...
-
 UAVCAN-based networks can accommodate up to 127 nodes on the same logical bus.
-More nodes can be added to the system by separating the bus into several independent logical segments
-interconnected via gateway nodes.
 
-UAVCAN supports up to 32768 distinct message data types and up to 256 distinct service data types.
-Part of those are reserved for the standard data types defined by the specification;
-the rest are available for vendor- and application-specific data types.
-More information is provided in the chapter \ref{sec:application_layer}.
+UAVCAN supports up to 65536 message, 256 service, and 8 anonymous data types.
```
pavel-kirienko: I suggest a different wording:

> and at least 8 anonymous data types.

"at least" works better because the transport layer spec says that the subset of message type ID usable with anonymous transfers is dependent on the transport layer in use.

specification/introduction/introduction.tex
```diff
@@ -8,73 +8,69 @@ \section{Design principles}
...
+Within each priority level different types of transfers and different data types are
+prioritized in a well-defined, deterministic manner.
 
-The list of transport protocols supported by UAVCAN is provided in the chapter \ref{sec:transport_layer}.
+The list of transport protocols supported by UAVCAN is provided in chapter \ref{sec:transport_layer}.
 Non-redundant, doubly-redundant and triply-redundant transports are supported.
 More information on the physical layer and standardized physical connectivity options
 is provided in the chapter \ref{sec:physical_layer}.
 
+\mbox{\textbf{TODO}: add more about standard application capabilities when this section is complete.}
```
pavel-kirienko: As the document is getting larger, we might be risking forgetting about this TODO. I would rather open a ticket.

specification/introduction/introduction.tex
```diff
@@ -8,73 +8,69 @@ \section{Design principles}
...
-Non-COTS applications may avoid any dependency on the standard data type definitions (barring the aforementioned
-minor exceptions), relying solely on vendor-specific data types instead.
+forum\footnote{Please refer to \href{http://uavcan.org}{uavcan.org}.}
+which is open to the general public.
+
+The set of standard data type definitions\footnote{The Data Structure Description Language (DSDL) and related concepts are described in section \ref{sec:dsdl}.}
+is an integral part of the specification, however, there is only a small set of required data types needed to implement the protocol.
+A larger set of optional data types are defined to create a standardized data exchange environment
+supporting the interoperability of COTS\footnote{Commercial off-the-shelf equipment.} equipment manufactured by different vendors.
+While the maintainers are charged with maintining interoperability they will also accommodate custom data types defined for closed systems to be used in place of the provided types.
```
pavel-kirienko: `maintining` (I use VSCode with `ban.spellright`, works great with LaTeX)

> While the maintainers are charged with maintining interoperability they will also accommodate custom data types defined for closed systems to be used in place of the provided types.

This might be misinterpreted as if we're to allow inclusion of application-specific or vendor-specific data types into the standard set. I suggest removing this sentence completely.

specification/introduction/introduction.tex
thirtytwobits: 👌

specification/introduction/introduction.tex
thirtytwobits: See #22

specification/introduction/introduction.tex
thirtytwobits: 👌

PR #23, opened 2018-09-12
Issue #16 edits to section 2.1

### LINKED ISSUE #4: Proposed: New High-Level Definition of UAVCAN

thirtytwobits:
This includes a proposal to close #4  if we can agree on this diagram.

### DIFF
--- a/specification/basic_concepts/basic_concepts.tex
+++ b/specification/basic_concepts/basic_concepts.tex
@@ -1,10 +1,10 @@
 \chapter{Basic concepts}\label{sec:basic_concepts}
 
-UAVCAN is a lightweight protocol designed to provide a highly reliable communication method for
-aerospace and robotic applications via robust vehicle bus networks.
+
 A UAVCAN network is a decentralized peer network, where each peer (node) has a unique
 numeric identifier --- \emph{node ID} --- ranging from 1 to 127, inclusively.
 Nodes of a UAVCAN network can communicate using the following communication methods:
+
 \begin{description}
     \item[Message broadcasting] --- The primary method of data exchange with one-to-all publish/subscribe semantics.
     \item[Service invocation] --- The communication method for peer-to-peer request/response
@@ -14,10 +14,10 @@ Nodes of a UAVCAN network can communicate using the following communication meth
 For each type of communication, a predefined set of data structures is used,
 where each data structure has a unique identifier --- the \emph{data type ID} (DTID).
 Additionally, every data structure definition has a pair of major and minor semantic version numbers,
-which enable data type definitions to evolve in arbitrary ways while ensuring a comprehensible
-migration path in the event of backward-incompatible changes.
-Some data structures are standard and defined by the protocol specification;
-others may be specific to a particular application or vendor.
+which enable data type definitions to evolve in arbitrary ways while ensuring a well-defined
+migration path if backward-incompatible changes are introduced.
+Some data structures are standard and defined by the protocol specification (of which only a small 
+subset are required); others may be specific to a particular application or vendor.
 
 Since every message or service data type has its own unique data type ID,
 and each node in the network has its own unique node ID,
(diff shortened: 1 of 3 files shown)


pavel-kirienko:
@thirtytwobits could you please attach a PDF here?

@kjetilkjeka should we expect the build bot to work or is it not quite there yet?

thirtytwobits:
[UAVCAN_Specification.pdf](https://github.com/UAVCAN/specification/files/2381791/UAVCAN_Specification.pdf)

Built on my mac. Not sure if my tex environment is correct but you should get the gist.

specification/basic_concepts/basic_concepts.tex
```diff
@@ -29,38 +29,30 @@ \chapter{Basic concepts}\label{sec:basic_concepts}
 A DSDL description can be used to automatically generate the serialization/deserialization code
 for every defined data structure in a particular programming language.
 DSDL ensures that the worst case memory footprint and computational complexity per data type
-are constant and easily predictable, which is paramount for hard real-time and safety-critical applications.
+are constant and easily predictable.
```
pavel-kirienko: I think the removed statement was beneficial for emphasizing that the protocol is specifically designed for these kinds of applications. Would you not agree?

I understand that the text might be redundant in some aspects, but that is probably good for clarity.

specification/basic_concepts/basic_concepts.tex
```diff
@@ -29,38 +29,30 @@ \chapter{Basic concepts}\label{sec:basic_concepts}
...
 On top of the standard data types, UAVCAN defines a set of standard high-level functions including:
-node health monitoring, network discovery, time synchronization, firmware update,
+node health monitoring, node discovery, time synchronization, firmware update,
 plug-and-play node support, and more.
 For more information see chapter \ref{sec:application_layer}.
 
 Serialized message and service data structures are exchanged by means of the transport
 layer (chapter \ref{sec:transport_layer}),
-which implements automatic decomposition/reassembly of long transfers into/from several transport
+which implements automatic decomposition/reassembly of long transfers from/into several transport
```
pavel-kirienko: > which implements automatic decomposition of long transfers into several transport frames

> which implements automatic reassembly of long transfers from several transport frames

Decomposition from; reassembly into. That seemed correct to me, although I could be missing something.

specification/basic_concepts/basic_concepts.tex
```diff
@@ -29,38 +29,30 @@ \chapter{Basic concepts}\label{sec:basic_concepts}
...
-
-        \hline
-        \multicolumn{4}{|c|}{Serialization} \\
-        \hline
-        \multicolumn{4}{|c|}{Transport layer} \\
+    \begin{tabular}{|c|c|l|c|l|}
         \hline
-    \end{tabularx}
+        \multicolumn{5}{|c|}{Application} \\ \hline
+        \multicolumn{1}{|l|}{\cellcolor[HTML]{FCE5CE}} & \cellcolor[HTML]{B5D8A6}{\color[HTML]{000000} Required Functions} & \cellcolor[HTML]{DD7E6B} & \cellcolor[HTML]{FFE593}{\color[HTML]{333333} Standard Functions} & \cellcolor[HTML]{D6A5BD} \\ \cline{2-2} \cline{4-4}
```
pavel-kirienko: Inconsistent capitalization: Title Case is not used in this diagram excepting the two cells you added: `Required Functions` and `Standard Functions`. I suggest replacing them with `Required functions` and `Standard functions`.

Also, at this point we might want to consider adding `Custom functions` for extra consistency?

Also also, not sure what went wrong, but the table is rendered black-and-white, despite the cell color directives. Perhaps they should be removed, seeing as they don't work? We can get back to visual refinements later when the actual content is finished, if desired.

specification/basic_concepts/basic_concepts.tex
thirtytwobits: This is a normative section so I've tried to remove anything that might be construed as "marketing". We should move all suggestions for suitability or other text promoting the standard's use to the non-normative "introduction" section. 

That aside, this text really is more a snippet from a embedded systems course and doesn't add information about how to implement UAVCAN. It's correct but superfluous.

specification/basic_concepts/basic_concepts.tex
thirtytwobits: We're using / here in place of 'or' so I was trying to order the sentence so decomposition would align with from and reassembly would align with into but now that I read it I think this is backwards. That sais, this sentence would make English majors riot so perhaps we should just avoid trickery and go with:

```
Serialized message and service data structures are exchanged by means of the transport
layer (chapter \ref{sec:transport_layer}), which implements automatic decomposition-into and reassembly-from long transfers to/from several transport frames\footnote{Here and elsewhere, a \emph{transport frame} means a block of data
that can be atomically exchanged over the network, e.g., a CAN frame.},
allowing nodes to exchange data structures of arbitrary size.
```

specification/basic_concepts/basic_concepts.tex
thirtytwobits: ...and this is where I admit I have no friggin idea how to write TeX :/ I used an online TeX table generator for this. I'll fix the issues you pointed out but I don't know why the colouring hasn't worked.

thirtytwobits:
[UAVCAN_Specification.pdf](https://github.com/UAVCAN/specification/files/2391205/UAVCAN_Specification.pdf)

specification/basic_concepts/basic_concepts.tex
```diff
@@ -29,38 +29,31 @@ \chapter{Basic concepts}\label{sec:basic_concepts}
...
-        \multicolumn{4}{|c|}{Transport layer} \\
+    \begin{tabular}{|c|c|l|c|l|c|}
         \hline
-    \end{tabularx}
+        \multicolumn{6}{|c|}{Applications}                                                                                              \\ \hline
+        \multicolumn{1}{|l|}{} & Required Functions &            & Standard Functions          &            & Custom Functions          \\ \cline{2-2} \cline{4-4} \cline{6-6} 
+        \multicolumn{2}{|c|}{Required Data Types}   & \multicolumn{2}{c|}{Standard Data Types} & \multicolumn{2}{c|}{Custom Data Types} \\ \hline
+        \multicolumn{6}{|c|}{Serialization}                                                                                             \\ \hline
+        \multicolumn{6}{|c|}{Transport}                                                                                                 \\ \hline
+    \end{tabular}
```
pavel-kirienko: Not to be nitpicking, but let us consider consistency issues. Everywhere else in the document, `Sentence case` is preferred rather than `Title Case`, including headings and tables. I'm not much into UX stuff, so correct me if I'm wrong, but Title Case seems to be a thing of the American flavor of English. Should we migrate the whole document to adopt that convention? I am weakly against that, and, frankly, I don't think I have much room to speak here as my knowledge of English is imperfect at best.

specification/introduction/introduction.tex
```diff
@@ -1,7 +1,11 @@
 \chapter{Introduction}\label{sec:introduction}
 
-This chapter covers the basic concepts that govern development and maintenance of the specification.
-The actual specification is contained in the following chapters.
+UAVCAN is a lightweight protocol designed to provide a highly reliable communication method for
+aerospace and robotic applications via robust vehicle bus networks.
+
+\begin{mdframed}
+    This is a non-nomative section covering the basic concepts that govern development and maintenance of the specification.
+\end{mdframed}
```
pavel-kirienko: Can we please get rid of the frame? ;) Also the `r` I mentioned earlier still seems to be missing.

specification/basic_concepts/basic_concepts.tex
thirtytwobits: Not a nit-pick at all. I'm always unsure which case to use in technical writing but the suggestion that is more "American" makes me agree that we should _not_ use it (The last thing we need is more bombast in this world). I'll fix.

Issue #8, opened 2018-08-02
Redundant interfaces


kjetilkjeka:
@pavel-kirienko I have a couple of questions regarding the redundant interfaces. 
 1. Do we have to require the users to let the first one time out before using the next one. Wouldn't an implementation where the interface where the message is received first on was used and then detection of a lost interface could be done even after the message was received on the "backup" interface.
 2. Do we need every node to use the same interface as number 1. and number 2. If we have nodes that didn't require redundancy wouldn't it be perfectly fine to split them on the two interfaces for load sharing reasons?

pavel-kirienko:
> Do we have to require the users to let the first one time out before using the next one. Wouldn't an implementation where the interface where the message is received first on was used and then detection of a lost interface could be done even after the message was received on the "backup" interface.

Dangerous. The transfer ID wraps around every 32 transfers. Different interfaces are expected to exhibit different latencies even on a properly functioning system, especially if the system shares redundantly-interfaced and non-redundantly-interfaced nodes. If the latency of a backup interface relative to the primary interface exceeds 32*`transfer_interval`, and we were to allow receiving nodes to switch between interfaces freely disregarding the timeout, the receiving node would skip the whole period of transfer ID (32 transfers will be lost). The problem primarily affects low-priority transfers where large latencies are possible.

> Do we need every node to use the same interface as number 1. and number 2. If we have nodes that didn't require redundancy wouldn't it be perfectly fine to split them on the two interfaces for load sharing reasons?

No. Redundant interfaces are there not for load-sharing reasons. Whenever you add a node to the bus you increase the likelihood of its failure, hence the backup bus (-es) should interconnect only mission-critical equipment.

![image](https://user-images.githubusercontent.com/3298404/43633789-eaa76948-9712-11e8-8317-a047a9b779a2.png)

PR #24, opened 2018-09-23
Note on the use of the transfer ID value for purposes of temporal synchronization of message streams

pavel-kirienko:
Added a minor footnote on the page 28 section 4.2.1. [UAVCAN_Specification.pdf](https://github.com/UAVCAN/specification/files/2408965/UAVCAN_Specification.pdf)

### DIFF
--- a/specification/transport_layer/transport_layer.tex
+++ b/specification/transport_layer/transport_layer.tex
@@ -386,7 +386,11 @@ It is recognized that perfectly simultaneous transmission may not be possible du
 utilization rates of the redundant interfaces and different phasing of their traffic;
 however, that is not an issue for UAVCAN.
 If perfectly simultaneous frame submission is not possible, interfaces with lower numerical index
-should be handled in the first order.
+should be handled in the first order\footnote{It should be kept in mind that redundant interfaces
+are used for increased fault tolerance, not for load sharing reasons.
+Whenever a node is connected to an interface, the likelihood of its failure is increased;
+therefore, backup interfaces should interconnect only mission-critical equipment.
+See section \ref{sec:phy_non_uniform_transport_redundancy}.}.
 
 An exception to the above rule applies if the payload of the transfer depends on some properties
 of the interface through which the transfer is emitted.
@@ -540,6 +544,16 @@ require any background maintenance processes.
     to the application requirements (e.g., maximum recovery time in the event of an interface failure),
     or the protocol stack can estimate this value automatically by analyzing the transfer intervals. \\
 \end{UAVCANSimpleTable}
+
+It might be useful to clarify why immediate switching between interfaces is not allowed,
+and why nodes are required to let the first interface time out before using the next one.
+The transfer ID value wraps around every 32 transfers.
+Different interfaces are expected to exhibit different latencies even in a properly functioning system,
+especially if the system contains both redundantly-interfaced and non-redundantly-interfaced nodes.
+If the latency of a backup interface relative to the primary interface exceeds 32 transfer intervals,
+and receiving nodes were to be allowed to switch between interfaces freely disregarding the timeout,
+the receiving node would skip the whole period of transfer ID (32 transfers will be lost).
+The problem would primarily affect low-priority transfers where large latencies are possible.
 
 \subsection{State update in a redundant interface configuration}
 
--- a/specification/transport_layer/transport_layer.tex
+++ b/specification/transport_layer/transport_layer.tex
@@ -178,9 +178,8 @@ that is provided for every transfer.
 This value is crucial for many aspects of UAVCAN communication\footnote{One might be tempted to use the transfer ID
 value for temporal synchronization of parallel message streams originating from the same node,
 where messages bearing the same transfer ID value are supposed to correspond to the same moment of time.
-Such use case is explicitly discouraged due to the fact that the overflow interval of the transfer ID value
-is insufficiently long, which creates the risk of undetected phase slip of one of the message streams by
-more than 32 messages behind.
+Such use is strongly discouraged because it is impossible to detect if one node is more than
+32 messages behind another.
 If temporal synchronization is necessary, explicit time stamping should be used instead.};
 specifically:
 \begin{description}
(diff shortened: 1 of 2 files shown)


pavel-kirienko:
Added two more clarifications that fix https://github.com/UAVCAN/specification/issues/8.

[UAVCAN_Specification.pdf](https://github.com/UAVCAN/specification/files/2408980/UAVCAN_Specification.pdf)

specification/transport_layer/transport_layer.tex
```diff
@@ -379,7 +386,11 @@ \subsection{Redundant interface support}
 utilization rates of the redundant interfaces and different phasing of their traffic;
 however, that is not an issue for UAVCAN.
 If perfectly simultaneous frame submission is not possible, interfaces with lower numerical index
-should be handled in the first order.
+should be handled in the first order\footnote{It should be kept in mind that redundant interfaces
+are used for increased fault tolerance, not for load sharing reasons.
+Whenever a node is connected to an interface, the likelihood of its failure is increased;
```
thirtytwobits: Comma splice and ambiguity in the predicate. Perhaps: Whenever a node is connected to an interface the likelihood of the interface failing is increased

specification/transport_layer/transport_layer.tex
```diff
@@ -379,7 +386,11 @@ \subsection{Redundant interface support}
 utilization rates of the redundant interfaces and different phasing of their traffic;
 however, that is not an issue for UAVCAN.
 If perfectly simultaneous frame submission is not possible, interfaces with lower numerical index
-should be handled in the first order.
+should be handled in the first order\footnote{It should be kept in mind that redundant interfaces
+are used for increased fault tolerance, not for load sharing reasons.
```
thirtytwobits: I think this first sentence is superfluous to the footnote and could be omitted.

specification/transport_layer/transport_layer.tex
```diff
@@ -379,7 +386,11 @@ \subsection{Redundant interface support}
...
+Whenever a node is connected to an interface, the likelihood of its failure is increased;
+therefore, backup interfaces should interconnect only mission-critical equipment.
```
thirtytwobits: This is a non-normative comment so we shouldn't use "should" here. Perhaps: This suggests that backup interfaces may only interconnect with mission-critical equipment.

I know there is a counter argument to be made in having asymmetrical redundant busses where it adds complexity because the testing matrix is increased (i.e. you have to prove that mission-critical component X can operate nominally on bus B where node Y,Z, and Q are no longer available). This even makes the wiring harnesses different per bus which is, perhaps, added complexity and does mean the bus capacitance has to be calculated differently. Considering there is very little that is not mission-critical on some drones (i.e. where your ESCs are the only thing on the bus) I wouldn't put too much weight on this footnote.

specification/transport_layer/transport_layer.tex
```diff
@@ -534,6 +545,16 @@ \subsection{State variables}
     or the protocol stack can estimate this value automatically by analyzing the transfer intervals. \\
 \end{UAVCANSimpleTable}
 
+It might be useful to clarify why immediate switching between interfaces is not allowed,
```
thirtytwobits: This is voiced conversationally whereas the rest of the specification is in "shall/should" voice. Could we adopt a box format to put non-normative and/or expositional text like this? It's jarring to be reading a spec and having a human suddenly take over the text and start talking.

specification/transport_layer/transport_layer.tex
kjetilkjeka: I think including explicitly that "redundant interfaces are used for increased fault tolerance, not for load sharing reasons.". can clear up misunderstandings and should be included in the text itself and not as a footnote. 

I don't think "kept in mind" is a very good phrasing since this hasn't been mentioned in the text before.

specification/transport_layer/transport_layer.tex
pavel-kirienko: > It's jarring to be reading a spec and having a human suddenly take over the text and start talking.

A very good point. I suggest just keeping the style consistent throughout rather than adopting the box convention you suggested. Fixed this in my latest commit.

pavel-kirienko:
I think I have addressed your concerns. I understand that my English is seriously lacking; however, I expect that @hamishwillee will be able to help us out on this when the first draft is nearly finished.

specification/transport_layer/transport_layer.tex
```diff
@@ -534,6 +545,15 @@ \subsection{State variables}
     or the protocol stack can estimate this value automatically by analyzing the transfer intervals. \\
 \end{UAVCANSimpleTable}
 
+The reason nodes are required to let the first interface time out before using the next one
```
thirtytwobits: One more English nit:

Nodes are required to let the first interface timeout before using the next one because the transfer ID field is expected to wrap around frequently...

specification/transport_layer/transport_layer.tex
```diff
@@ -534,6 +545,15 @@ \subsection{State variables}
...
+is that the transfer ID field is expected to wrap around frequently (every 32 transfers).
+Different interfaces are expected to exhibit different latencies even in a properly functioning system,
+especially if the system contains both redundantly-interfaced and non-redundantly-interfaced nodes.
+If the latency of a backup interface relative to the primary interface exceeds 32 transfer intervals,
+and receiving nodes were to be allowed to switch between interfaces freely disregarding the timeout,
+the receiving node would skip the whole period of transfer ID (32 transfers will be lost).
+The problem would primarily affect low-priority transfers where large latencies are possible.
```
thirtytwobits: s/affect/effect/

specification/transport_layer/transport_layer.tex
```diff
@@ -534,6 +545,15 @@ \subsection{State variables}
     or the protocol stack can estimate this value automatically by analyzing the transfer intervals. \\
 \end{UAVCANSimpleTable}
 
+The reason nodes are required to let the first interface time out before using the next one
+is that the transfer ID field is expected to wrap around frequently (every 32 transfers).
+Different interfaces are expected to exhibit different latencies even in a properly functioning system,
+especially if the system contains both redundantly-interfaced and non-redundantly-interfaced nodes.
+If the latency of a backup interface relative to the primary interface exceeds 32 transfer intervals,
+and receiving nodes were to be allowed to switch between interfaces freely disregarding the timeout,
+the receiving node would skip the whole period of transfer ID (32 transfers will be lost).
```
thirtytwobits: ...the whole period of transfer IDs...

specification/transport_layer/transport_layer.tex
pavel-kirienko: According to https://www.grammarly.com/blog/affect-vs-effect/ that would be a mistake.

PR #31, opened 2018-10-08
Landscape page CAN ID table

kjetilkjeka:
This PR put the CAN ID table on a landscape page in a similar style that was used on the web page docs. This makes it possible to both include names and ranges in the table. I find this new table easier to read and much more useful in itself (i.g. you can see that the Anonymous Node ID is 0 directly from this table).

While I was at it I updated the fields to the layout we talked about at the Stockholm summit.

### DIFF
--- a/specification/transport_layer/can.tex
+++ b/specification/transport_layer/can.tex
@@ -24,10 +24,10 @@ summarize the purpose of the fields and their permitted values
 for message transfers, anonymous message transfers, and service transfers, respectively.
 
 
-\begin{landscape}
 % Please do not remove the hard placement specifier [H], it is needed to keep elements ordered.
 \begin{figure}[H]
 \centering
+\resizebox{\textwidth}{!}{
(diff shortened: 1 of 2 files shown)


kjetilkjeka:
I forgot that the automatic building didn't work on PR from forks, I'll reopen this PR from a branch

pavel-kirienko:
While you're at it, consider limiting the line width to 120 characters please.

kjetilkjeka:
> I forgot that the automatic building didn't work on PR from forks, I'll reopen this PR from a branch

Looks like it will build the commit if it exists on the upstream repo as well

pavel-kirienko:
I was about to start updating the document myself; as the scope of the required changes is quite extensive, it would be best if everyone else refrained from introducing further changes over the next few days. I will merge this changeset into my branch though, as soon as it's in the master.

pavel-kirienko:
How necessary is it to use the landscape page orientation here? Looks quite awkward. Can we not squeeze everything into a portrait-oriented page, perhaps by reducing the font size or margins?

kjetilkjeka:
My only strong opinion is that the horizontal kind of table is more readable and useful than the vertical type.

I resized the table to fit a pagewidth and removed the landscape page. Is this better?

PR #32, opened 2018-11-11
Chapters 1 and 2

pavel-kirienko:
The first two chapters are virtually finished. I merely updated them according to our recent discussions, nothing new here. Note that the master branch is currently a staging branch, so the changes are not to be considered final after merging.

[UAVCAN_Specification.pdf](https://github.com/UAVCAN/specification/files/2569794/UAVCAN_Specification.pdf)

### DIFF
--- a/specification/basic_concepts/basic_concepts.tex
+++ b/specification/basic_concepts/basic_concepts.tex
@@ -28,7 +28,7 @@ subset are required); others may be specific to a particular application or vend
 
 \subsubsection{Subjects and services}
 
-Message exchanges between nodes are grouped into \emph{subjects} by their semantic meaning.
+Message exchanges between nodes are grouped into \emph{subjects} by the semantic meaning of the message.
 Message exchanges belonging to the same subject use same message data type down to the major version
 (minor versions are allowed to differ) and pertain to same function or process within the system.
 
--- a/specification/basic_concepts/basic_concepts.tex
+++ b/specification/basic_concepts/basic_concepts.tex
@@ -48,7 +48,7 @@ service data types contain two data structures: one for request, and the other f
 If a port identifier is defined by a data type definition, such identifier is said to be a
 \emph{regulated port identifier}
 (by inheritance, the same naming principle applies to \emph{regulated subject ID} and \emph{regulated service ID}).
-If a regulated port ID is not provided with a data type definition,
+If a data type definition is not provided with a regulated port ID,
 it can be assigned at runtime on a per-application basis,
 in which case it is said to be an \emph{unregulated port identifier}.
 A data type can be used with an arbitrary number of different unregulated port identifiers,
--- a/specification/basic_concepts/basic_concepts.tex
+++ b/specification/basic_concepts/basic_concepts.tex
@@ -93,7 +93,7 @@ since that is likely to cause unresolvable port identifier collisions\footnote{
     from said external party since the designers of the system do not have control over their fixed port identifiers.
     Because of this, the specification strongly discourages the use of fixed unregulated private port identifiers.
     If a data type definition is ever disclosed to any other party (i.e., a party that did not author it)
-    or to the public at large, such data type should be devoid of a fixed port identifier.
+    or to the public at large it is important that the data type \emph{not} include a fixed port-identifier.
 }.
 This restriction shall be followed at all times by all compliant implementations and
 systems\footnote{
--- a/specification/basic_concepts/basic_concepts.tex
+++ b/specification/basic_concepts/basic_concepts.tex
@@ -63,21 +63,22 @@ but not more than one fixed port identifier.
 
 \subsubsection{Regulation}
 
-Data type definitions can be created by the UAVCAN specification maintainers or by third parties,
-such as vendors or application designers.
+Data type definitions can be created by the UAVCAN specification maintainers or by its users,
+such as equipment vendors or application designers.
 Irrespective of the origin, data types can be included into the set of data type definitions maintained
 and distributed by the UAVCAN specification maintainers;
 definitions belonging to this set are termed \emph{regulated data type definitions}.
 The specification maintainers undertake to keep regulated definitions well-maintained and may occasionally
-amend them and release new versions, if such actions are believed to benefit the ecosystem.
-Data type definitions created by third parties that are not included into the aforementioned set are called
-\emph{unregulated data type definitions}.
+amend them and release new versions, if such actions are believed to benefit the protocol.
+User-created (i.e., vendor-specific or application-specific) data type definitions that are
+not included into the aforementioned set are called \emph{unregulated data type definitions}.
 
 Unregulated definitions that are made available for reuse by others are called
 \emph{unregulated public data type definitions};
-those that are kept closed-source for private use by third parties that created them are called
+those that are kept closed-source for private use by their authors are called
 \emph{(unregulated) private data type definitions}\footnote{The word \emph{``unregulated''} is redundant
-because private data types cannot be regulated, by definition.}.
+because private data types cannot be regulated, by definition.
+Likewise, all regulated definitions are public, so the word \emph{``public''} can be omitted.}.
 
 Data type definitions authored by the specification maintainers for the purpose of supporting and advancing
 this specification are called \emph{standard data type definitions}.
@@ -85,56 +86,62 @@ All standard data type definitions are regulated.
 
 Fixed port identifiers can be used only with regulated data type definitions or with private definitions.
 Fixed port identifiers must not be used with public unregulated data types,
-since that is likely to cause unresolvable port identifier collisions.
-This restriction must be followed at all times by all parties that use the specification;
-if a private data type definition is ever disclosed to any other party (i.e., a party that did not author it)
-or to the public at large, such data type \emph{must be devoid of a fixed port identifier}.
-Equipment and systems that fail to follow this requirement will not be considered UAVCAN-compliant.
+since that is likely to cause unresolvable port identifier collisions\footnote{
+    Any system that relies on data type definitions with fixed port identifiers provided by
+    an external party (i.e., the data types and the system in question are designed by different parties)
+    runs the risk of encountering port identifier conflicts that cannot be resolved without resorting to help
+    of such external party, since the designers of the system do not have control of their fixed port identifiers.
+    This is why the specification strongly discourages the use of fixed unregulated private port identifiers.
+    If a data type definition is ever disclosed to any other party (i.e., a party that did not author it)
+    or to the public at large, such data type should be devoid of a fixed port identifier.
+}.
+This restriction shall be followed at all times by all compliant implementations and
+systems\footnote{
+    In general, private unregulated fixed port identifiers are collision-prone by their nature, so they should
+    be avoided unless there are very strong reasons for their usage and the authors fully understand the risks.
+}.
 
 \begin{UAVCANSimpleTable}{Data type taxonomy}{|l|X|X|}
     & Regulated & Unregulated \\
     \bfseries{Public}
     &
-    Standard and third-party definitions.\newline
-    Fixed port ID are allowed; they are called \emph{``regulated port ID''}.
+    Standard and contributed (e.g., vendor-specific) definitions.\newline
+    Fixed port identifiers are allowed; they are called \emph{``regulated port ID''}.
     &
-    Third-party definitions distributed separately from the UAVCAN specification.\newline
-    Fixed port ID are \emph{not allowed}.
+    Definitions distributed separately from the UAVCAN specification.\newline
+    Fixed port identifiers are \emph{not allowed}.
     \\
 
     \bfseries{Private}
     &
     Nonexistent category.
     &
-    Third-party definitions that are not available to anyone except the party that authored them.\newline
-    Fixed port ID are permitted as long as the definition is only used by the party that created it;
-    they are called \emph{``unregulated fixed port ID''}.
+    Definitions that are typically not available to anyone except their authors.\newline
+    Fixed port identifiers are permitted (although not recommended) as long as the definition is only used by
+    its authors; they are called \emph{``unregulated fixed port ID''}.
     \\
 \end{UAVCANSimpleTable}
 
+DSDL processing tools shall prohibit unregulated fixed port identifiers by default,
+unless they are explicitly configured otherwise.
+
 Each of the two sets of valid port identifiers (which are subject identifiers and port identifiers) are
 segregated into three categories: application-specific (i.e., non-fixed or unregulated) identifiers,
-regulated non-standard port identifiers (i.e., data type definitions submitted by third parties),
-and standard port identifiers.
+regulated non-standard port identifiers, and standard port identifiers.
 Fixed unregulated port identifiers can be allocated in the same range where application-specific identifiers are.
 The ranges are documented in chapter \ref{sec:application_layer}.
 
-DSDL processing tools are recommended to prohibit unregulated fixed port identifiers by default,
-unless they are explicitly configured otherwise.
-
-Third party authors that desire to release regulated data type definitions or contribute to the standard data
+Data type authors that desire to release regulated data type definitions or contribute to the standard data
 type set should contact the UAVCAN maintainers for coordination.
 The maintainers will choose unoccupied fixed port identifiers for use with the new definitions, if necessary.
 Since the set of regulated definitions is maintained in a highly centralized manner,
 it can be statically ensured that no identifier collisions will take place within it;
 also, since the identifier ranges used with regulated definitions are segregated from the rest,
-it is also guaranteed that no regulated port ID will ever be in conflict with any other compliant UAVCAN
-node or system.
-
-The motivation for the prohibition of fixed port identifiers in third-party unregulated public data types is
-derived directly from the above:
-since there is no central repository of unregulated definitions,
-collisions would be likely to happen, which would be extremely damaging to the ecosystem.
+regulated port ID will not conflict with any other compliant UAVCAN node or system\footnote{
+    The motivation for the prohibition of fixed port identifiers in unregulated public data types is
+    derived directly from the above: since there is no central repository of unregulated definitions,
+    collisions would be likely to happen, which would be extremely damaging to the ecosystem.
+}.
 
 \subsubsection{Serialization}
 
@@ -165,7 +172,9 @@ For more information see chapter \ref{sec:application_layer}.
         \hline
         \multicolumn{6}{|c|}{Applications} \\ \hline
 
-        \multicolumn{1}{|l|}{} & Required functions & & Standard functions & & Custom functions \\
+        \qquad{} & Required functions &
+        \qquad{} & Standard functions &
+        \qquad{} & Custom functions \\
         \cline{2-2} \cline{4-4} \cline{6-6}
 
         \multicolumn{2}{|c|}{Required data types} &
--- a/specification/basic_concepts/basic_concepts.tex
+++ b/specification/basic_concepts/basic_concepts.tex
@@ -32,7 +32,8 @@ Message exchanges between nodes are grouped into \emph{subjects} by the semantic
 Message exchanges belonging to the same subject use same message data type down to the major version
 (minor versions are allowed to differ) and pertain to same function or process within the system.
 
-Request/response exchanges between nodes are grouped into \emph{services} by their semantic meaning,
+Request/response exchanges between nodes are grouped into \emph{services} by the semantic meaning
+of the request and response,
 like messages are grouped into subjects.
 Requests and their corresponding responses that belong to the same service use same service data type down to
 the major version (minor versions are allowed to differ; as a consequence, the minor data type version number
--- a/specification/basic_concepts/basic_concepts.tex
+++ b/specification/basic_concepts/basic_concepts.tex
@@ -2,6 +2,10 @@
 
 \section{Main principles}
 
+\subsection{Communication}
+
+\subsubsection{Architecture}
+
 A UAVCAN network is a decentralized peer network, where each peer (node) has a unique
 numeric identifier\footnote{Here and elsewhere in this specification, \emph{ID} and \emph{identifier} are used
 interchangeably unless specifically indicated otherwise.}
@@ -22,6 +26,8 @@ migration path if backward-incompatible changes are introduced.
 Some data types are standard and defined by the protocol specification (of which only a small
 subset are required); others may be specific to a particular application or vendor.
 
+\subsubsection{Subjects and services}
+
 Message exchanges between nodes are grouped into \emph{subjects} by their semantic meaning.
 Message exchanges belonging to the same subject use same message data type down to the major version
 (minor versions are allowed to differ) and pertain to same function or process within the system.
@@ -38,6 +44,10 @@ An umbrella term \emph{port ID} is used to refer either to a subject ID or to a 
 (port identifiers have no direct manifestation in the construction of the protocol,
 but they are convenient for discussion).
 
+\subsection{Data types}
+
+\subsubsection{Data type definitions}
+
 Message and service data types\footnote{Data types are the core concept of UAVCAN.}
 are defined using the \emph{data structure description language} (DSDL) (chapter \ref{sec:dsdl}).
 A DSDL definition specifies the name, major version, minor version, the data structure,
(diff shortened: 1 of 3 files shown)


pavel-kirienko:
@kjetilkjeka @thirtytwobits I am wondering if we can somehow expedite the review here as it is almost holding me back right now as I am working on chapter 4 (transport). I realize that the sections pertaining to data type versioning may require updates later depending on the resolution of the ongoing discussion about versioning, but that probably should not prevent this PR from being merged.

specification/introduction/introduction.tex
```diff
@@ -3,59 +3,78 @@ \chapter{Introduction}\label{sec:introduction}
...
     \item[Democratic network] --- There will be no master node.
     All nodes in the network will have the same communication rights; there must be no single point of failure.
 
-    \item[Nodes can exchange long payloads] --- Nodes must be provided with a simple way of exchanging large
-    data structures that cannot fit into a single transport frame (such as GNSS solutions, 3D vectors, etc.).
-    UAVCAN should perform automatic transfer decomposition and reassembly at the protocol level,
+    \item[High-level communication abstractions] --- The protocol will support publish/subscribe and remote procedure
+    call communication semantics with statically defined and statically verified data types (schema).
+    The data types used for communication will be defined in a clear, platform-agnostic way
+    that can be easily understood by machines, including humans.
```
thirtytwobits: did you mean to call humans "machines" in this sentence?

specification/introduction/introduction.tex
```diff
@@ -64,31 +83,30 @@ \section{Maintenance of the standard data type set}
...
+related concepts are described in section \ref{sec:dsdl}.} is an integral part of the specification;
+however, there is only a small set of required data types needed to implement the protocol.
 A larger set of optional data types are defined to create a standardized data exchange environment
-supporting the interoperability of COTS\footnote{Commercial off-the-shelf equipment.} equipment manufactured by different vendors.
+supporting the interoperability of COTS\footnote{Commercial off-the-shelf equipment.}
+equipment manufactured by different vendors.
 See chapter \ref{sec:application_layer} for more information.
 
-Within the same major version, the set of standard data type definitions can be modified only in the following ways:
+Within the same major version of the specification,
```
thirtytwobits: Should we add a similar section for what modifications are allowed within a minor version?

specification/introduction/introduction.tex
pavel-kirienko: I should probably fix this otherwise some adepts of free will might come after me. For now, let's pretend this never happened.

specification/basic_concepts/basic_concepts.tex
```diff
@@ -1,33 +1,62 @@
...
-For each type of communication, a predefined set of data structures is used,
-where each data structure has a unique identifier --- the \emph{data type ID} (DTID).
-Additionally, every data structure definition has a pair of major and minor semantic version numbers,
+For each type of communication, a predefined set of data types is used,
+where each data type has a unique name.
+Additionally, every data type definition has a pair of major and minor version numbers,
 which enable data type definitions to evolve in arbitrary ways while ensuring a well-defined
 migration path if backward-incompatible changes are introduced.
-Some data structures are standard and defined by the protocol specification (of which only a small 
+Some data types are standard and defined by the protocol specification (of which only a small
```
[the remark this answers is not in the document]
pavel-kirienko: What's being said here is that some types are defined by the "protocol specification", not just "protocol". We could rephrase it to state "defined by the specification" if that adds clarity. I don't think we can just throw in a word "library" without explaining what it is exactly (this term does not appear anywhere in the specification currently, we only introduce "root namespaces" as the closest thing to a library).

specification/introduction/introduction.tex
pavel-kirienko: Let us finish the DSDL chapter first, and then we'll consider whether the whole section "Maintenance of the standard data type set" should be moved into the DSDL chapter. For now I suggest to keep it as-is.

specification/basic_concepts/basic_concepts.tex
```diff
@@ -1,33 +1,62 @@
...
+are defined using the \emph{data structure description language} (DSDL) (chapter \ref{sec:dsdl}).
+A DSDL definition specifies the name, major version, minor version, the data structure,
+and an optional predefined port ID of the data type.
+Message data types always contain exactly one data structure, whereas
+service data types contain two data structures: one for request, and the other for response.
+
+If a port identifier is defined by a data type definition, such identifier is said to be a
+\emph{regulated port identifier}
+(by inheritance, the same naming principle applies to \emph{regulated subject ID} and \emph{regulated service ID}).
+If a regulated port ID is not provided with a data type definition,
```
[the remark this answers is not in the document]
pavel-kirienko: Ah, I see how it can be misread, indeed. What I meant is "if a regulated port ID is not specified for a data type definition". Will change as you suggested.

specification/basic_concepts/basic_concepts.tex
```diff
@@ -1,33 +1,62 @@
...
+
+Message and service data types\footnote{Data types are the core concept of UAVCAN.}
+are defined using the \emph{data structure description language} (DSDL) (chapter \ref{sec:dsdl}).
+A DSDL definition specifies the name, major version, minor version, the data structure,
+and an optional predefined port ID of the data type.
+Message data types always contain exactly one data structure, whereas
+service data types contain two data structures: one for request, and the other for response.
+
+If a port identifier is defined by a data type definition, such identifier is said to be a
+\emph{regulated port identifier}
```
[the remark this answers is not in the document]
pavel-kirienko: The name is per this discussion; my understanding was that we've all agreed with (or at least tolerated) the word "regulated": https://forum.uavcan.org/t/on-subject-service-id-ranges/195/4?u=pavel.kirienko

specification/basic_concepts/basic_concepts.tex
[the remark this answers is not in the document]
kjetilkjeka: I see the problem. What I'm worried about is that the sentence may mislead readers to expect to find the standard types defined inside the protocol specification document. I'm happy with any rephrasing that will make the core protocol a bit more seperated from the standard dsdl.

Perhaps it would work out to say "There exists standard types that defines....." leaving the explaining of how/where they are defined till later.

specification/basic_concepts/basic_concepts.tex
```diff
@@ -1,33 +1,62 @@
...
+(by inheritance, the same naming principle applies to \emph{regulated subject ID} and \emph{regulated service ID}).
+If a regulated port ID is not provided with a data type definition,
+it can be assigned at runtime on a per-application basis,
+in which case it is said to be an \emph{unregulated port identifier}.
```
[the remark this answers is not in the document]
pavel-kirienko: "Standard" is probably a poor fit because we also use this word to describe specification-defined data types ("standard data types"). In this discussion, we have agreed to use separate ranges for regulated and application-specific identifiers, hence the split you mentioned: https://forum.uavcan.org/t/on-subject-service-id-ranges/195/4?u=pavel.kirienko

specification/introduction/introduction.tex
```diff
@@ -3,59 +3,81 @@ \chapter{Introduction}\label{sec:introduction}
...
+    node software update, node status monitoring\footnote{Which naturally grows into a vehicle-wide
+    health monitoring solution.}, network-wide time synchronization, plug-and-play node support, etc.
+
+    \item[Atomic data abstractions] --- Nodes must be provided with a simple way of exchanging large
+    data structures that exceed the capacity of a single transport frame\footnote{A \emph{transport frame} is
+    an atomic transmission unit defined by the underlying transport protocol. For example, a CAN frame.}.
+    UAVCAN should perform automatic data decomposition and reassembly at the protocol level,
     hiding the related complexity from the application.
 
+    \item[High throughput, low latency, determinism] --- UAVCAN must be suitable for use in
```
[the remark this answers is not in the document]
pavel-kirienko: > This is a bit "Yes we want them alls"

The listed properties are not mutually exclusive so I don't think there is a problem. Do you have a better wording in mind? We shouldn't mention CAN specifically because we are now officially transport-agnostic.

specification/basic_concepts/basic_concepts.tex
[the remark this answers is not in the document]
pavel-kirienko: > What I'm worried about is that the sentence may mislead readers to expect to find the standard types defined inside the protocol specification document.

Ooh, I was actually considering to do just that, but still not sure if it makes sense. We could fully automate the task, too. This should be discussed when we get closer to Chapter 5 "Application layer".

specification/basic_concepts/basic_concepts.tex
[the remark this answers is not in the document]
pavel-kirienko: Fixed

specification/introduction/introduction.tex
```diff
@@ -3,59 +3,81 @@ \chapter{Introduction}\label{sec:introduction}
...
 
 \section{Capabilities}
 
 UAVCAN-based networks can accommodate up to 127 nodes on the same logical bus.
 
-UAVCAN supports up to 65536 message, 256 service, and at least 8 anonymous data types.
-Some of these identifiers are reserved for the core standard with
-the rest available for vendor- and application-specific data types.
+UAVCAN supports an unlimited number of data types, which can be either standard (defined by the specification)
+or vendor-specific (defined by third parties for private use or for public release).
```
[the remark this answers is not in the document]
pavel-kirienko: Fixed?

specification/introduction/introduction.tex
```diff
@@ -3,59 +3,81 @@ \chapter{Introduction}\label{sec:introduction}
...
+    \item[High throughput, low latency, determinism] --- UAVCAN must be suitable for use in
+    hard real-time deterministic systems.
+
     \item[Support for redundant interfaces and redundant nodes] --- UAVCAN must be suitable for use in
     applications that require modular redundancy.
 
-    \item[High throughput, low latency communication] --- UAVCAN must be suitable for use in
-    hard real-time systems.
-
     \item[Simple logic, low computational requirements] --- UAVCAN targets a wide variety of embedded systems,
```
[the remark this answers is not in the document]
pavel-kirienko: Consider this a form of marketing. A human engineer would open the doc and read through the list of goals while evaluating each entry against their own problems they have to solve. I think it is important to make a very obvious statement that we do, in fact, care about memory impaired microcontrollers because the preceding entries might lead one to believe otherwise.

specification/basic_concepts/basic_concepts.tex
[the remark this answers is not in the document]
kjetilkjeka: No it's quite different.

How you present it in the text is if I create an application specific definition with a standard suject ID of 11, 11 would be a regulated port identifier. And if IDs allocated for the standard types do not have a type which sets them as their standard ID they're unregulated.

How I expected it to be is that regulated IDs are IDs which is globally vetted for conflicts, while unregulated is done per application.

specification/basic_concepts/basic_concepts.tex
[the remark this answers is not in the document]
kjetilkjeka: How you define regulated in the text is not how it's used in that discussion. See my previous comment.

specification/introduction/introduction.tex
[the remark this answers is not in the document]
kjetilkjeka: :+1:

specification/introduction/introduction.tex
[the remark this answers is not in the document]
kjetilkjeka: My comment was meant as an addition. Also meant as "marketing". Stating that supporting the minimal set of the protocol will be inexpensive in terms of engineering hours and change required, and more advanced support can be incrementally added in a pick and choose fashion as more features is desired.

specification/basic_concepts/basic_concepts.tex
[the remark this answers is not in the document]
kjetilkjeka: I would avoid doing this for any other types than the most essential:
 - NodeStatus/Heartbeat is required
 - GetNodeInfo is nice to do it for
 - GetPortInfo (and friends?) is nice to do it for

I think doing it for file, diagnostics, etc would be a bit messy since directory structure (namespacing) is not preserved well in a spec sheet.

specification/basic_concepts/basic_concepts.tex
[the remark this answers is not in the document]
pavel-kirienko: > How you present it in the text is if I create an application specific definition with a standard suject ID of 11, 11 would be a regulated port identifier.

Yes, this is the intended meaning of the text (and your definition would be non-compliant and rejected by parsers). The background is presented in the first post here https://forum.uavcan.org/t/on-subject-service-id-ranges/195, specifically:

> The above has an implication that vendors can no longer define private DTS that contain static Subject/Service IDs, becase if that were the case, definitions from different vendors might conflict with each other and there would be no way of fixing that without modifying the DTS, because, see, the IDs are unchangeable. Public DTS, on the other hand, can rely on static ID as long as they are all centralized in one place, where the lack of collisions can be determined statically. Such centralized maintenance can be achieved by either assigning ID ranges to vendors (which seems wasteful, as the ID spaces are limited), or by keeping all vendor-specific definitions (in their respective namespaces) in one central repository, such as https://github.com/UAVCAN/vendor-specific-dsdl.
>
> If the above reasoning is found sound, we are to end up with the following kinds of DTS:
>
> - Standard DTS, where identifiers can be static (conflicts are not possible because the set is the same for everyone) or dynamic.
> - Vendor-specific public DTS, where the identifiers can be static (conflicts are avoidable if we keep all definitions in one central repository or assign per-vendor ranges) or dynamic.
> - Vendor-specific private DTS, where the identifiers can be only dynamic, because vendors wouldn’t be able to guarantee collision-free coexistence of their data types.

To be more specific, you can't really define a data type with a regulated subject ID set to 11, because it lies outside the regulated ranges we discussed. PyDSDL checks for that specifically and throws an error.

This might seem restrictive, as you won't be able to define your own private application-specific data types with fixed identifiers, but this is necessary to statically guarantee collision-free coexistence of nodes from different vendors that use private data types. The core idea is that if you want a fixed ID, it must be public, so that other vendors can see it and avoid reusing it.

specification/introduction/introduction.tex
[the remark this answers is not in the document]
pavel-kirienko: Fixed?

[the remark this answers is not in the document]
kjetilkjeka: :+1:

specification/basic_concepts/basic_concepts.tex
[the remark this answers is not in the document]
kjetilkjeka: I thought this sorted out by the "this range can also be used with static IDS post" later on. By rejecting definitions of default service/subject ID in the application specific domain will make the life of groups with full control of the nodes harder without gaining any security as conflicts may be configured dynamically anyway.

> This might seem restrictive, as you won't be able to define your own private application-specific data types with fixed identifiers, but this is necessary to statically guarantee collision-free coexistence of nodes from different vendors that use private data types. The core idea is that if you want a fixed ID, it must be public, so that other vendors can see it and avoid reusing it.

When nodes from different vendors use application-specific data types they will be configured dynamically. At the end of the day, the integrator will have control of what Port IDs they use, and can avoid using the same Port IDs allocated for other things (both dynamically and with the default port ID for custom messages).

On the other hand, checking that no definitions in the vendor specific (or standard) repo use these IDs are essential.

specification/introduction/introduction.tex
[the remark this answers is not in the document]
kjetilkjeka: It's hard without mentioning CAN. What about: "Uavcan adds very little overhead to the transport layer, giving it high throughput while retaining low latency and suitability for hard real time usage."

specification/basic_concepts/basic_concepts.tex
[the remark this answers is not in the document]
pavel-kirienko: > By rejecting definitions of default service/subject ID in the application specific domain will make the life of groups with full control of the nodes harder

This is true, but if we were to lift the restriction, such groups would have to remain committed to being fully isolated from the rest of the ecosystem once they started using private types with fixed ID, otherwise addition of any third-party (e.g. COTS) unit can potentially cause a conflict. Knowing that addition of foreign third-party hardware may cause a conflict will encourage divergence and fragmentation of the ecosystem, which we must avoid. We might sort of work around this by defining complex rules, such as "if a device is intended to be used outside of the vendor's ecosystem, it shall not use fixed identifiers in the non-regulated range", but I don't expect such a rule to work well in practice, if at all (because it would be difficult to enforce automatically, and humans are bad at following rules).

We might come up with some clever solutions to this later in UAVCAN v1.x. Remember, it is easier to lift restrictions than to introduce them later in a backward compatible way.

> At the end of the day, the integrator will have control of what Port IDs they use, and can avoid using the same Port IDs allocated for other things (both dynamically and with the default port ID for custom messages).

As per our discussion, a fixed port ID is a guarantee that the data type will always be accessible at this port ID. It is possible to use it with different ports as well, though. Context: https://forum.uavcan.org/t/data-type-compatibility-assurance/196/7?u=pavel.kirienko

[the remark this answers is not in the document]
kjetilkjeka: > This is true, but if we were to lift the restriction, such groups would have to remain committed to being fully isolated from the rest of the ecosystem once they started using private types with fixed ID, otherwise addition of any third-party (e.g. COTS) unit can potentially cause a conflict.

This is not true. Any third party unit would either use the vendor specific DTS or be configurable inside the application specific range. They would always be able to configure the new unit to IDs they're not using. If not, any COTS unit might also cause conflict with COTS units from a different brand.

If they were starting to sell the unit (instead of just the integrated application) they would have to move the definition to the vendor specific domain and request a new static ID (or use dynamically configured IDs). They would have to move their definition to the vendor specific namespace either way (to make it available and check out on Port compatibility checks) which would be impossible with an ID from the application specific DTS.

> We might sort of work around this by defining complex rules, such as "if a device is intended to be used outside of the vendor's ecosystem, it shall not use fixed identifiers in the non-regulated range", but I don't expect such a rule to work well in practice, if at all (because it would be difficult to enforce automatically, and humans are bad at following rules).

We only need one simple rule. The application integrator always decides the IDs inside the application DTS. If it's impossible to assign default IDs for types that only make sense to be used in one single application people are going to work around it by defining the ID as an associated constant, which results in the same problem + frustration.

specification/introduction/introduction.tex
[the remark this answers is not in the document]
kjetilkjeka: :+1:

pavel-kirienko:
@kjetilkjeka @thirtytwobits updated per discussion on the forum: https://forum.uavcan.org/t/on-standards-and-regulation/260

pavel-kirienko:
Thanks, waiting for @thirtytwobits

pavel-kirienko:
I see that Scott has approved this PR, but the change I submitted today is quite substantial, so I believe it needs another review.

specification/basic_concepts/basic_concepts.tex
```diff
@@ -38,33 +44,106 @@ \section{Main principles}
...
+
+\begin{UAVCANSimpleTable}{Data type taxonomy}{|l|X|X|}
+    & Regulated & Unregulated \\
+    \bfseries{Public}
+    &
+    Standard and third-party definitions.\newline
+    Fixed port ID are allowed; they are called \emph{``regulated port ID''}.
+    &
+    Third-party definitions distributed separately from the UAVCAN specification.\newline
+    Fixed port ID are \emph{not allowed}.
```
thirtytwobits: %s/Fixed port ID are/Fixed port IDs are/g

specification/basic_concepts/basic_concepts.tex
```diff
@@ -38,33 +44,106 @@ \section{Main principles}
...
+Equipment and systems that fail to follow this requirement will not be considered UAVCAN-compliant.
+
+\begin{UAVCANSimpleTable}{Data type taxonomy}{|l|X|X|}
+    & Regulated & Unregulated \\
+    \bfseries{Public}
+    &
+    Standard and third-party definitions.\newline
+    Fixed port ID are allowed; they are called \emph{``regulated port ID''}.
+    &
+    Third-party definitions distributed separately from the UAVCAN specification.\newline
```
thirtytwobits: The use of the term "Third-party" makes no sense within this specification given that it is not a contract (i.e. there are no parties). I think we need to say "Definitions not distributed with the UAVCAN specification" here and use a similar strategy for the rest of this section.

specification/basic_concepts/basic_concepts.tex
```diff
@@ -38,33 +44,106 @@ \section{Main principles}
...
+\end{UAVCANSimpleTable}
+
+Each of the two sets of valid port identifiers (which are subject identifiers and port identifiers) are
+segregated into three categories: application-specific (i.e., non-fixed or unregulated) identifiers,
+regulated non-standard port identifiers (i.e., data type definitions submitted by third parties),
+and standard port identifiers.
+Fixed unregulated port identifiers can be allocated in the same range where application-specific identifiers are.
+The ranges are documented in chapter \ref{sec:application_layer}.
+
+DSDL processing tools are recommended to prohibit unregulated fixed port identifiers by default,
```
thirtytwobits: %s/are recommended to/shall

We might as well make this a normative requirement for DSDL processing. Ambiguity doesn't provide much on this point.

specification/basic_concepts/basic_concepts.tex
```diff
@@ -38,33 +44,106 @@ \section{Main principles}
...
+unless they are explicitly configured otherwise.
+
+Third party authors that desire to release regulated data type definitions or contribute to the standard data
+type set should contact the UAVCAN maintainers for coordination.
+The maintainers will choose unoccupied fixed port identifiers for use with the new definitions, if necessary.
+Since the set of regulated definitions is maintained in a highly centralized manner,
+it can be statically ensured that no identifier collisions will take place within it;
+also, since the identifier ranges used with regulated definitions are segregated from the rest,
+it is also guaranteed that no regulated port ID will ever be in conflict with any other compliant UAVCAN
```
thirtytwobits: No guarantees, no warranties. We need to soften this to be non-normative language that details how we can help the ecosystem do the right thing; until we don't. I.e. if you (Pavel) get hit by a bus and nobody maintains the identifier registry there is no legal entity that is liable for the collapse of this mechanism. "Use this help at your own risk." We're not even liable if we get it wrong and accidentally re-issue an already issued identifier. Please remember that while we discuss "maintainers" this role does not legally exist and maintainers cannot not provide any real services.

specification/basic_concepts/basic_concepts.tex
```diff
@@ -38,33 +44,106 @@ \section{Main principles}
...
+node or system.
+
+The motivation for the prohibition of fixed port identifiers in third-party unregulated public data types is
+derived directly from the above:
+since there is no central repository of unregulated definitions,
+collisions would be likely to happen, which would be extremely damaging to the ecosystem.
```
thirtytwobits: The specification should not discuss ecosystems in any normative section. An ecosystem is something that is completely outside of the purview of this specification.

pavel-kirienko:
@thirtytwobits ready for review. The wording is still a bit clumsy and the text could use some restructuring; I suggest we postpone that for another pass after the other chapters are finished.

specification/basic_concepts/basic_concepts.tex
```diff
@@ -1,79 +1,209 @@
...
-and each node in the network has its own unique node ID,
-a pair of data type ID and node ID can be used to support redundant nodes with identical
-functionality inside the same network.
+\subsubsection{Subjects and services}
+
+Message exchanges between nodes are grouped into \emph{subjects} by their semantic meaning.
+Message exchanges belonging to the same subject use same message data type down to the major version
+(minor versions are allowed to differ) and pertain to same function or process within the system.
+
+Request/response exchanges between nodes are grouped into \emph{services} by their semantic meaning,
```
thirtytwobits: same comment here about "their" being ambiguous

specification/basic_concepts/basic_concepts.tex
```diff
@@ -1,79 +1,209 @@
...
+like messages are grouped into subjects.
+Requests and their corresponding responses that belong to the same service use same service data type down to
+the major version (minor versions are allowed to differ; as a consequence, the minor data type version number
+of a service response may differ from that of its corresponding request) and pertain to same function.
```
thirtytwobits:
```suggestion
of a service response may differ from that of its corresponding request) and pertain to the same function.
```

specification/basic_concepts/basic_concepts.tex
```diff
@@ -1,79 +1,209 @@
...
+An umbrella term \emph{port ID} is used to refer either to a subject ID or to a service ID
+(port identifiers have no direct manifestation in the construction of the protocol,
+but they are convenient for discussion).
+
+\subsection{Data types}
+
+\subsubsection{Data type definitions}
+
+Message and service data types\footnote{Data types are the core concept of UAVCAN.}
+are defined using the \emph{data structure description language} (DSDL) (chapter \ref{sec:dsdl}).
```
thirtytwobits:
```suggestion
are defined using the \emph{Data Structure Description Language} (DSDL) (chapter \ref{sec:dsdl}).
```

specification/basic_concepts/basic_concepts.tex
```diff
@@ -1,79 +1,209 @@
...
+A DSDL definition specifies the name, major version, minor version, the data structure,
+and an optional predefined port ID of the data type.
```
thirtytwobits: When I first encountered "data type ids are optional" it really threw me for a loop. I'd love to get footnote or link here helping the reader understand that port IDs are eventually required but can be resolved later by a system when not using predefined identifiers. It's redundant and non-normative to mention it at this point but a footnote would help the read understand that the port ID is not, ultimately, optional.

specification/basic_concepts/basic_concepts.tex
```diff
@@ -1,79 +1,209 @@
...
+this specification are called \emph{standard data type definitions}.
+All standard data type definitions are regulated.
+
+Fixed port identifiers can be used only with regulated data type definitions or with private definitions.
+Fixed port identifiers must not be used with public unregulated data types,
+since that is likely to cause unresolvable port identifier collisions\footnote{
+    Any system that relies on data type definitions with fixed port identifiers provided by
+    an external party (i.e., data types and the system in question are designed by different parties)
+    runs the risk of encountering port identifier conflicts that cannot be resolved without resorting to help
+    of such external party, since the designers of the system do not have control over their fixed port identifiers.
```
thirtytwobits:
```suggestion
    from said external party since the designers of the system do not have control over their fixed port identifiers.
```

specification/basic_concepts/basic_concepts.tex
```diff
@@ -1,79 +1,209 @@
...
+    This is why the specification strongly discourages the use of fixed unregulated private port identifiers.
```
thirtytwobits:
```suggestion
    Because of this, the specification strongly discourages the use of fixed unregulated private port identifiers.
```

specification/basic_concepts/basic_concepts.tex
```diff
@@ -1,79 +1,209 @@
...
+    If a data type definition is ever disclosed to any other party (i.e., a party that did not author it)
+    or to the public at large, such data type should be devoid of a fixed port identifier.
```
thirtytwobits:
```suggestion
    or to the public at large it is important that the data type \emph{not} include a fixed port-identifier.
```

specification/basic_concepts/basic_concepts.tex
```diff
@@ -1,79 +1,209 @@
...
+DSDL processing tools shall prohibit unregulated fixed port identifiers by default,
+unless they are explicitly configured otherwise.
+
+Each of the two sets of valid port identifiers (which are subject identifiers and port identifiers) are
+segregated into three categories: application-specific (i.e., non-fixed or unregulated) identifiers,
+regulated non-standard port identifiers, and standard port identifiers.
+Fixed unregulated port identifiers can be allocated in the same range where application-specific identifiers are.
+The ranges are documented in chapter \ref{sec:application_layer}.
+
+Data type authors that desire to release regulated data type definitions or contribute to the standard data
```
thirtytwobits:
```suggestion
Data type authors that want to release regulated data type definitions or contribute to the standard data
```

specification/basic_concepts/basic_concepts.tex
```diff
@@ -1,79 +1,209 @@
...
+type set should contact the UAVCAN maintainers for coordination.
+The maintainers will choose unoccupied fixed port identifiers for use with the new definitions, if necessary.
+Since the set of regulated definitions is maintained in a highly centralized manner,
+it can be statically ensured that no identifier collisions will take place within it;
+also, since the identifier ranges used with regulated definitions are segregated from the rest,
```
thirtytwobits:
```suggestion
also, since the identifier ranges used with regulated definitions are segregated,
```

specification/basic_concepts/basic_concepts.tex
```diff
@@ -1,79 +1,209 @@
...
+regulated port ID will not conflict with any other compliant UAVCAN node or system\footnote{
```
thirtytwobits:
```suggestion
regulated port-IDs will not conflict with any other compliant UAVCAN node or system\footnote{
```
In general I'd use a hyphen when discussing types of port identifiers to be explicit about what term modifies which (e.g. is the port or the identifier regulated: "regulated-port identifier" or "regulated port-identifier")

specification/basic_concepts/basic_concepts.tex
```diff
@@ -1,79 +1,209 @@
...
+    The motivation for the prohibition of fixed port identifiers in unregulated public data types is
+    derived directly from the above: since there is no central repository of unregulated definitions,
+    collisions would be likely to happen, which would be extremely damaging to the ecosystem.
```
thirtytwobits:
```suggestion
    collisions would be likely.
```

There is no ecosystem that is regulated by this specification.

specification/introduction/introduction.tex
```diff
@@ -3,59 +3,83 @@ \chapter{Introduction}\label{sec:introduction}
...
+    \item[Facilitation of functional safety] --- A system designer relying on UAVCAN will have the necessary
+    guarantees and tools at their disposal to analyze the system and ensure its correct behavior.
+
+    \item[High-level communication abstractions] --- The protocol will support publish/subscribe and remote procedure
+    call communication semantics with statically defined and statically verified data types (schema).
+    The data types used for communication will be defined in a clear, platform-agnostic way
+    that can be easily understood by machines, including humans.
+
+    \item[Facilitation of cross-vendor interoperability] --- UAVCAN will be a common foundation that
+    different vendors can build upon to ensure interoperability of their equipment.
```
thirtytwobits:
```suggestion
    different vendors can build upon to maximize interoperability of their equipment.
```
no assurances are provided by the specification.

specification/introduction/introduction.tex
```diff
@@ -3,59 +3,83 @@ \chapter{Introduction}\label{sec:introduction}
...
-    node software update, node status monitoring (which naturally grows into a vehicle-wide health monitoring),
-    network-wide time synchronization, dynamic node ID allocation (a.k.a. plug-and-play node support), etc.
+    \item[Support for various transport protocols] --- UAVCAN will be usable with different transports.
+    The standard must be capable of accommodating other transport protocols in the future.
 
-    \item[Open specification and reference implementations] --- The UAVCAN specification is open and
-    freely available; the reference implementations are distributed under the terms of the permissive MIT License.
+    \item[Open specification and reference implementations] --- The UAVCAN specification will always be open and
+    free to use for everyone; the reference implementations will be distributed under the terms of
+    the permissive MIT License or released into public domain.
```
thirtytwobits:
```suggestion
    the permissive MIT License or released into the public domain.
```

specification/introduction/introduction.tex
```diff
@@ -64,31 +88,30 @@ \section{Maintenance of the standard data type set}
...
-    \end{itemize}
-
-    \item An existing data type can be declared deprecated.
+    \item An existing data type or a particular major version of it can be declared deprecated.
     \begin{itemize}
         \item Once declared deprecated, the data type will be maintained for at least two more years.
-        After this period its default data type ID may be reused for an incompatible data type.
+        After this period, its regulated port ID (if defined) may be reused for an incompatible data type definition.
+        The maintainers will be striving to postpone the reuse of regulated port identifiers as much as
+        possible in order to eliminate the possibility of unintended conflicts.
```
thirtytwobits:
```suggestion
        possible in order to minimize the possibility of unintended conflicts.
```
no warrenties

specification/basic_concepts/basic_concepts.tex
pavel-kirienko: This is not supposed to be a proper noun, why capitalize?

pavel-kirienko: Let's get back to this during the final proofreading.

specification/basic_concepts/basic_concepts.tex
pavel-kirienko: They are not required for nested types. I am going to move this question into a separate ticket.

PR #33, opened 2018-11-16
DSDL constification + assert

kjetilkjeka:
Opening this PR early to expose some details.

### DIFF
--- a/specification/dsdl/dsdl.tex
+++ b/specification/dsdl/dsdl.tex
@@ -134,11 +134,11 @@ not crossing the boundary of the service response marker.
 
 
 \subsection{Constant expressions}\label{sec:dsdl_constant_expressions}
-A valid constant expression is either just a constant, or an expression that results in a constant.
-Such expression is given in infix notation,
-this means \texttt{A <op> B} for binary operators and \texttt{<op> A} for unary operators.
-A comprehensive table of legal operations on types is defined below,
-the order of the table also defines the order of precedence between operations.
+A valid constant expression is either a constant literal, or an expression that evaluates to a constant.
+The legal constant literals can be found in Section \ref{sec:dsdl_primitive_constant_data_types},
+and a comprehensive table of legal operations and their precedence is given below.
+Infix notation is used, meaning a constant evaluation is written as \texttt{A <op> B} for 
+binary operators and \texttt{<op> A} for unary operators, where \texttt{A} and \texttt{B} are constant expressions.
 
 \begin{UAVCANSimpleTable}{Operators for constant expressions}{|c c c c c l|} \label{table:dsdl_const_operators}
 Precedence & Op & A & B & Res & Description \\
@@ -380,15 +380,14 @@ if the value is \verb|false| the resulting type must be considered invalid, and 
 Both checking the assertions before compilation and compiling the assertions into compile time assertions in the
 resulting languages is considered legal way to handle this as long as it results in no way to use the resulting type.
 
-Keywords that are interesting in relation to assertions are given below.
-To allow both earlier described ways of assertions checking,
-they are designed in a way such that a DSDL compiler can substitute them for their value early on.
-This allow all assertion checking to be done at a separate pass before compilation.
-They can also be compiled into values used in target language compile time assertions in the resulting type definition.
+There exists a few magical keywords only available for assertions.
+This is a deliberate choice made to allow implementation of constant evaluation
+without needing the in-depth analysis required when checking assertions.
+The keywords and their described behavior follows: 
 
 \begin{UAVCANSimpleTable}{Compatibility between associated const and const type}{|l l|}
 \textbf{Keyword} & \textbf{Description} \\
-\texttt{offset}  & The set of possible bit offsets when serializing.
+\texttt{offset} & Evaluates into the set of possible bit offsets at the point of evaluation.
 \end{UAVCANSimpleTable}
 
 \subsection{Comments}
--- a/specification/dsdl/dsdl.tex
+++ b/specification/dsdl/dsdl.tex
@@ -473,28 +473,28 @@ All the primitive const types are given below together with the literal used for
 
     \footnotesize \texttt{const int} &
         \footnotesize $\supseteq \lbrack -\frac{2^{64}}{2}, 2^{64} - 1\rbrack$ &
-        \footnotesize \texttt{0 |
+        \footnotesize \verb~0 |
                             (\textbackslash+|-)?[1-9][0-9]* |
                             (\textbackslash+|-)?0x([0-9]|[a-f])+ |
                             (\textbackslash+|-)?0o[0-7]+ |
-                            (\textbackslash+|-)?0b(1|2)+ } \\
+                            (\textbackslash+|-)?0b(1|2)+~ \\
 
     \footnotesize \texttt{const float} &
         \footnotesize \texttt{IEEE 754-2008 (binary64)} &
-        \footnotesize \texttt{(\textbackslash+|-)?[0-9]*.[0-9]* |
+        \footnotesize \verb~(\textbackslash+|-)?[0-9]*.[0-9]* |
                             (\textbackslash+|-)?[0-9]*((e|E)-?[0-9]+) |
-                            (\textbackslash+|-)?[0-9]*.[0-9]*((e|E)-?[0-9]+) } \\
+                            (\textbackslash+|-)?[0-9]*.[0-9]*((e|E)-?[0-9]+)~ \\
 
     \footnotesize \texttt{const char} &
         \footnotesize $\lbrack0, 127\rbrack$ &
         \footnotesize
-            \texttt{ '[\^{}'\textbackslash\textbackslash]' |
+            \verb~'[\^{}'\textbackslash\textbackslash]' |
                 '\textbackslash x[0-9]+' |
                 '\textbackslash n' |
                 '\textbackslash r' |
                 '\textbackslash\textbackslash' |
                 '\textbackslash t' |
-                '\textbackslash ''} \\
+                '\textbackslash ''~ \\
 
 \end{UAVCANSimpleTable}
 
--- a/specification/dsdl/dsdl.tex
+++ b/specification/dsdl/dsdl.tex
@@ -405,7 +405,7 @@ All the primitive const types are given below together with the literal used for
         \footnotesize \texttt{true | false} \\
 
     \footnotesize \texttt{const int} &
-        \footnotesize $\supseteq \lbrack -\frac{2^{64}}{2}, 2^64 - 1\rbrack$ &
+        \footnotesize $\supseteq \lbrack -\frac{2^{64}}{2}, 2^{64} - 1\rbrack$ &
         \footnotesize \texttt{0 |
                             (\textbackslash+|-)?[1-9][0-9]* |
                             (\textbackslash+|-)?0x([0-9]|[a-f])+ |
--- a/specification/dsdl/dsdl.tex
+++ b/specification/dsdl/dsdl.tex
@@ -132,6 +132,29 @@ Unless specifically stated otherwise,
 directives apply only to the part of the service type definition where they are defined,
 not crossing the boundary of the service response marker.
 
+\subsection{Keywords}\label{sec:dsdl_keywords}
+A comprehensive list of non context sensitive keywords and a short description is given below, 
+a bold face \texttt{\textbf{X}} is interchangeable with a decimal number of one or more digits (\texttt{[0-9]+}).
+
+\begin{UAVCANSimpleTable}{A comprehensive table of DSDL keywords}{|l l|}
+\textbf{Keyword} & \textbf{Description} \\
+\texttt{saturated} & Cast mode. \\
+\texttt{truncated} & Cast mode. \\
+\texttt{offset} & Usable only in assertions, evaluates into the set of possible bit offsets. \\
+\texttt{int\textbf{X}} & Primitive type. \\
+\texttt{uint\textbf{X}} & Primitive type. \\
+\texttt{void\textbf{X}} & Primitive type. \\
+\texttt{bool} & Primitive type. \\
+\texttt{true} & \texttt{bool} variant. \\
+\texttt{false} & \texttt{bool} variant. \\
+\texttt{con} & Reserved. \\
+\texttt{prn} & Reserved. \\
+\texttt{aux} & Reserved. \\
+\texttt{nul} & Reserved. \\
+\texttt{com\textbf{X}} & Reserved. \\
+\texttt{lpt\textbf{X}} & Reserved. \\
+\end{UAVCANSimpleTable}
+
 
 \subsection{Constant expressions}\label{sec:dsdl_constant_expressions}
 A valid constant expression is either a constant literal, or an expression that evaluates to a constant.
@@ -382,7 +405,10 @@ resulting languages is considered legal way to handle this as long as it results
 
 There exists a few magical keywords only available for assertions.
 This is a deliberate choice made to allow implementation of constant evaluation
-without needing the in-depth analysis required when checking assertions.
+without needing the in-depth anal\begin{UAVCANSimpleTable}{Compatibility between associated const and const type}{|l l|}
+\textbf{Keyword} & \textbf{Description} \\
+\texttt{offset} & Evaluates into the set of possible bit offsets at the point of evaluation.
+\end{UAVCANSimpleTable}ysis required when checking assertions.
 The keywords and their described behavior follows: 
 
 \begin{UAVCANSimpleTable}{Compatibility between associated const and const type}{|l l|}
@@ -563,6 +589,7 @@ and relies on external validation of assertions, can choose not to implement the
 Field names, constant names, and type names must contain only ASCII alphanumeric characters and underscores
 \verb|[A-Za-z0-9_]|,
 and must begin with an ASCII alphabetic character \verb|[A-Za-z]|.
+These identifiers are also not allowed to be equal to any of the keywords listed in Section \ref{sec:dsdl_keywords}.
 Violation of this rule must be detected by the DSDL compiler and treated as a fatal error.
 
 \subsection{Optional}
--- a/specification/dsdl/dsdl.tex
+++ b/specification/dsdl/dsdl.tex
@@ -369,6 +369,28 @@ As explained in the section \ref{sec:dsdl_versioning},
 there are circumstances when the DSDL compiler is required to assume implicit deprecation,
 as if the \verb|@deprecated| directive was provided in the definition source.
 
+\subsubsection{Assertions}
+
+Keyword: \verb|@assert <constant expression>|.
+
+This directive checks a property that must hold for compilation to succeed.
+The provided constant expression must evaluate into a value to type \verb|const bool|.
+If the value is \verb|true| the assertion check will pass and the resulting type should be considered valid,
+if the value is \verb|false| the resulting type must be considered invalid, and steps must be taken to avoid usage.
+Both checking the assertions before compilation and compiling the assertions into compile time assertions in the
+resulting languages is considered legal way to handle this as long as it results in no way to use the resulting type.
+
+Keywords that are interesting in relation to assertions are given below.
+To allow both earlier described ways of assertions checking,
+they are designed in a way such that a DSDL compiler can substitute them for their value early on.
+This allow all assertion checking to be done at a separate pass before compilation.
+They can also be compiled into values used in target language compile time assertions in the resulting type definition.
+
+\begin{UAVCANSimpleTable}{Compatibility between associated const and const type}{|l l|}
+\textbf{Keyword} & \textbf{Description} \\
+\texttt{offset}  & The set of possible bit offsets when serializing.
+\end{UAVCANSimpleTable}
+
 \subsection{Comments}
 
 A DSDL description may contain comments starting from the ASCII number sign (\verb|#|)
--- a/specification/dsdl/dsdl.tex
+++ b/specification/dsdl/dsdl.tex
@@ -132,6 +132,51 @@ Unless specifically stated otherwise,
 directives apply only to the part of the service type definition where they are defined,
 not crossing the boundary of the service response marker.
 
+
+\subsection{Constant expressions}\label{sec:dsdl_constant_expressions}
+A valid constant expression is either just a constant, or an expression that results in a constant.
+Such expression is given in infix notation,
+this means \texttt{A <op> B} for binary operators and \texttt{<op> A} for unary operators.
+A comprehensive table of legal operations on types is defined below,
+the order of the table also defines the order of precedence between operations.
+
+\begin{UAVCANSimpleTable}{Operators for constant expressions}{|c c c c c l|} \label{table:dsdl_const_operators}
+Presedence & Op & A & B & Res & Description \\
+1 & \texttt{!}    &  \texttt{const bool}   & N/A                  & \texttt{const bool}    & Logical not \\
+
+
+2 & \texttt{*}    & \texttt{const int}    & \texttt{const int}   & \texttt{const int}     & Integer multiplication \\
+2 & \texttt{/}    & \texttt{const int}    & \texttt{const int}   & \texttt{const int}     & Integer division \\
+2 & \texttt{\%}   & \texttt{const int}    & \texttt{const int}   & \texttt{const int}     & Integer remainder \\
+
+2 & \texttt{*}    & \texttt{const float}  & \texttt{const float} & \texttt{const float}   & IEE754 multiplication \\
+2 & \texttt{/}    & \texttt{const float}  & \texttt{const float} & \texttt{const float}   & IEE754 division \\
+
+2 & \texttt{\%} & \texttt{\{const int\}}  & \texttt{const int}   & \texttt{\{const int\}} & Element-wise int remainder\\
+
+
+3 & \texttt{+}    & \texttt{const float}  & \texttt{const float} & \texttt{const float}   & IEEE754 sum \\
+3 & \texttt{-}    & \texttt{const float}  & \texttt{const float} & \texttt{const float}   & IEEE754 difference \\
+
+3 & \texttt{+}    & \texttt{const int}    & \texttt{const int}   & \texttt{const int}     & Integer sum \\
+3 & \texttt{-}    & \texttt{const int}    & \texttt{const int}   & \texttt{const int}     & Integer difference \\
+
+
+4 & \texttt{==}   & \texttt{const int}    & \texttt{const int}   & \texttt{const bool}    & Integer equality \\
+4 & \texttt{!=}   & \texttt{const int}    & \texttt{const int}   & \texttt{const bool}    & Integer inequality \\
+4 & \texttt{==}   & \texttt{const bool}   & \texttt{const bool}  & \texttt{const bool}    & Boolean equality \\
+4 & \texttt{!=}   & \texttt{const bool}   & \texttt{const bool}  & \texttt{const bool}    & Boolean inequality \\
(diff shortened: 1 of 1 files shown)


pavel-kirienko:
<sub>can we please limit the line length to 120 characters</sub>

kjetilkjeka:
@pavel-kirienko Hope these changes makes sense.

With the new const system:
 - Const definitions was renamed to associated const definitions. These constants are associated with a type defintion. And it hopefully it makes things less confusing when differentiating between constants used for things as defining array lengths and assertions and associated constant definitions.
 - There are no uint, int is expected to be big enough to be used for any uint or int. This is done to make type resolution from literals easy.

Next steps from here are:
 - Adding the const set type
 - Adding const evaluation (defining how basic operators needed for assert works)
 - Adding the assert which takes a `const bool` (and `offset` will be replaced by a set before const eval)

No need for a review beyond looking at the things I've stated, this is a work in progress PR at the moment.

pavel-kirienko:
> It is strictly defined what type of const literals that can be used to create what kind of associated consts.

Is PyDSDL currently conformant? https://github.com/UAVCAN/pydsdl/blob/9672256f4d57bc519d4ca5ebe82f13ae6e1b6c8d/pydsdl/data_type.py#L563-L606

specification/dsdl/dsdl.tex
```diff
@@ -132,30 +132,77 @@ \section{Syntax}\label{sec:dsdl_syntax}
...
-Field definitions follow one of the below specified declaration patterns:
+\subsection{Constant expressions}\label{sec:dsdl_constant_expressions}
+A valid constant expression is either just a constant, or an expression that results in a constant.
+Such expression is given in infix notation,
+this means \texttt{A <op> B} for binary operators and \texttt{<op> A} for unary operators.
+A comprehensive table of legal operations on types is defined below,
+the order of the table also defines the order of precedence between operations.
+
+\begin{UAVCANSimpleTable}{Operators for constant expressions}{|c c c c c l|} \label{table:dsdl_const_operators}
+Presedence & Op & A & B & Res & Description \\
```
pavel-kirienko: `Precedence`

If you are using vscode, I highly recommend https://marketplace.visualstudio.com/items?itemName=ban.spellright

specification/dsdl/dsdl.tex
```diff
@@ -132,30 +132,77 @@ \section{Syntax}\label{sec:dsdl_syntax}
...
+1 & \texttt{!}    &  \texttt{const bool}   & N/A                  & \texttt{const bool}    & Logical not \\
```
pavel-kirienko: Should we not use `not` instead? We are not encumbered with C legacy, and it makes sense to optimize for readability because DSDL types are write-once, read-anywhere kind of thing?

specification/dsdl/dsdl.tex
```diff
@@ -132,30 +132,77 @@ \section{Syntax}\label{sec:dsdl_syntax}
...
+4 & \texttt{==}   & \texttt{const bool}   & \texttt{const bool}  & \texttt{const bool}    & Boolean equality \\
+4 & \texttt{!=}   & \texttt{const bool}   & \texttt{const bool}  & \texttt{const bool}    & Boolean inequality \\
+4 & \texttt{==}   & \texttt{const char}   & \texttt{const char}  & \texttt{const bool}    & Character equality \\
+4 & \texttt{!=}   & \texttt{const char}   & \texttt{const char}  & \texttt{const bool}    & Character inequality \\
+4 & \texttt{==}   & \texttt{\{T\}}        & \texttt{\{T\}}       & \texttt{const bool}    & Set equality \\
+4 & \texttt{!=}   & \texttt{\{T\}}        & \texttt{\{T\}}       & \texttt{const bool}    & Set inequality \\
+
+
+5 & \texttt{\&\&} & \texttt{const bool}   & \texttt{const bool}  & \texttt{const bool}    & Logical and \\
+5 & \texttt{\&\&} & \texttt{const bool}   & \texttt{const bool}  & \texttt{const bool}    & Logical or
```
pavel-kirienko: Are we sure we want to continue the way of C and not use arguably more readable and possibly somehwat less error-prone alternatives of `and` and `or`? Also, typo: `&&` is stated both for "and" and "or".

specification/dsdl/dsdl.tex
```diff
@@ -376,6 +431,108 @@ \section{Primitive data types}\label{sec:dsdl_primitive_data_types}
...
+        \footnotesize $\{true, false\}$ &
+        \footnotesize \texttt{true | false} \\
+
+    \footnotesize \texttt{const int} &
+        \footnotesize $\supseteq \lbrack -\frac{2^{64}}{2}, 2^{64} - 1\rbrack$ &
+        \footnotesize \texttt{0 |
+                            (\textbackslash+|-)?[1-9][0-9]* |
+                            (\textbackslash+|-)?0x([0-9]|[a-f])+ |
+                            (\textbackslash+|-)?0o[0-7]+ |
+                            (\textbackslash+|-)?0b(1|2)+ } \\
```
pavel-kirienko: `0b(0|1)`

specification/dsdl/dsdl.tex
```diff
@@ -376,6 +431,108 @@ \section{Primitive data types}\label{sec:dsdl_primitive_data_types}
...
+        \footnotesize \texttt{IEEE 754-2008 (binary64)} &
+        \footnotesize \texttt{(\textbackslash+|-)?[0-9]*.[0-9]* |
+                            \textbackslash+|-)?[0-9]*((e|E)-?[0-9]+) |
+                            \textbackslash+|-)?[0-9]*.[0-9]*((e|E)-?[0-9]+) } \\
+
+    \footnotesize \texttt{const char} &
+        \footnotesize $\lbrack0, 127\rbrack$ &
+        \footnotesize
+            \texttt{ '[\^{}'\textbackslash\textbackslash]' |
+                'x[0-9]+' |
```
pavel-kirienko: The conventional form is `'\\x[0-9]+'`, is it not true? The leading backslash seems to be missing?

specification/dsdl/dsdl.tex
```diff
@@ -376,6 +431,108 @@ \section{Primitive data types}\label{sec:dsdl_primitive_data_types}
...
+                            (\textbackslash+|-)?[1-9][0-9]* |
+                            (\textbackslash+|-)?0x([0-9]|[a-f])+ |
+                            (\textbackslash+|-)?0o[0-7]+ |
+                            (\textbackslash+|-)?0b(1|2)+ } \\
+
+    \footnotesize \texttt{const float} &
+        \footnotesize \texttt{IEEE 754-2008 (binary64)} &
+        \footnotesize \texttt{(\textbackslash+|-)?[0-9]*.[0-9]* |
+                            \textbackslash+|-)?[0-9]*((e|E)-?[0-9]+) |
+                            \textbackslash+|-)?[0-9]*.[0-9]*((e|E)-?[0-9]+) } \\
```
pavel-kirienko: Mismatching parens.

specification/dsdl/dsdl.tex
```diff
@@ -376,6 +431,108 @@ \section{Primitive data types}\label{sec:dsdl_primitive_data_types}
...
 
+\subsection{Primitive constant data types}\label{sec:dsdl_primitive_constant_data_types}
+
+In addition to the ordinary data types that are used for defining types to be transmitted over the wire,
+there also exists data types only used for constant evaluation.
+The const data types are used for things that can be calculated only with the information self contained in the dsdl,
+such as defining array lengths, assertions and constant definitions.
+All the primitive const types are given below together with the literal used for construction.
+
+\begin{UAVCANSimpleTable}{Primitive constant data types}{|l l l|}
```
pavel-kirienko: I urge you to make the regex patterns formatted using `\verb||`, otherwise they just don't look right. Also, the table doesn't fit on the page (too wide); perhaps a list or some other form should be preferred?

specification/dsdl/dsdl.tex
```diff
@@ -376,6 +431,108 @@ \section{Primitive data types}\label{sec:dsdl_primitive_data_types}
...
+making it currently impossible to express the empty set in a convenient way as
+there would be no way to know the type of \verb|{}|.
+
+Examples of legal literals to construct sets are the following:
+\begin{itemize}
+    \item \texttt{ \{ 1, 2, 3 \} }
+    \item \texttt{ \{ 1\%3, 2\%3, 3\%3 \} }
+    \item \texttt{ \{ 1, 2, 3 , 3, 3, 3, 3\} }
+    \item \texttt{ \{ 0x201, 0b2, -301 \} }
+    \item \texttt{ \{ true, false, true \} }
```
pavel-kirienko: What non-integer sets are good for? None of the defined operators seem to yield them.

specification/dsdl/dsdl.tex
kjetilkjeka: The argument for using symbols over words is: it's pretty clear that symbols are not identifiers
 and If we want to add more at a later point, this can be done without reserving keywords prematurely.

Most modern language (that I'm aware of) still uses `!`, `&&` and `||`. The best reason to go against this would be to do whatever Pythons doing. But, Python being Python, replacing the symbols with the words should be trivial. i almost expect the following statement to parse `replace "!" with "not" in DSDL`.

specification/dsdl/dsdl.tex
kjetilkjeka: Thanks, willfix

specification/dsdl/dsdl.tex
kjetilkjeka: Thanks, willfix

specification/dsdl/dsdl.tex
kjetilkjeka: I can make another attempt, I generally had a lot of problems with `\verb` inside `UavcanSimpleTable` and started to avoid `\verb` outside normal text.

specification/dsdl/dsdl.tex
pavel-kirienko: I mean is a question of design. We should not design things to match our existing implementations. Surely we can find a way to evaluate an expression in Python, regardless of the chosen syntax. Shall we vote?

specification/dsdl/dsdl.tex
kjetilkjeka: It just seemed more future proof to define sets on things that can be checked equivalence on rather than integers only. 

The only things you can do with sets of `const bool`s or `const char`s is to construct them, and you're not even allowed to construct sets of `const float`s. The things is that constructing a const set is not a legal statement (on its own, it must be assigned to something or used in an assertion) in DSDL, and there is no legal place they can be used. This means that any statement that constructs such set will be erroneous in DSDL even if the construction itself has meaning. 

I can make it clear in the text that constructing sets of other stuff is technically legal but there is no way it will not give an error down the line and must be avoided.

It's totally OK for a compiler to give an error already on the construction of these sets as they will never succeed one step down the line.

specification/dsdl/dsdl.tex
kjetilkjeka: Thanks, willfix

specification/dsdl/dsdl.tex
kjetilkjeka: See comment above.

specification/dsdl/dsdl.tex
kjetilkjeka: Thanks, I will look into it.

specification/dsdl/dsdl.tex
pavel-kirienko: I mean what's the point of defining something that cannot exist? Shall the need arise in the future, we can always add them later in a backward-compatible way ¯\\_(ツ)_/¯

specification/dsdl/dsdl.tex
kjetilkjeka: > Shall we vote

Absolutely, +1 for symbols (`&&`, `||` and `!`) to be in line with C like syntax (used by: C/C++, C#, Java, Rust, etc) and, more importantly, to keep the number of reserved keywords low.

specification/dsdl/dsdl.tex
kjetilkjeka: It's to give a reason to disallow constructing the empty set `{}` without any considerations to type. This must be disallowed if we're to add them later.

I will add a paragraph to disallow this instead.

specification/dsdl/dsdl.tex
pavel-kirienko: I believe `verb` can be used with any separator, not necessarily `|`. How about `\verb~`?

kjetilkjeka: That worked

kjetilkjeka:
We should probably sort out the following before i complete this chapter:
 - DSDL versioning (https://forum.uavcan.org/t/simpler-and-better-dsdl-versioning/280/2)
 - How do make these boxes (https://forum.uavcan.org/t/non-normative-sections-of-the-specification/245/3)

pavel-kirienko:
The boxes should wait. That's a purely cosmetic change, we can postpone it until the text is in place.

kjetilkjeka:
The info is pretty much there.

A condensation of the core definitions is needed, this will be possible when non-normative boxes are available.

pavel-kirienko:
> A condensation of the core definitions is needed, this will be possible when non-normative boxes are available.

Consider merging this in: https://github.com/UAVCAN/specification/pull/40

pavel-kirienko:
@kjetilkjeka perhaps it's time to fix https://github.com/UAVCAN/specification/issues/34?

pavel-kirienko:
I just realized that this PR seems to be focused exclusively on const and expressions (yes I just read the title). In which case my suggestion to introduce other stuff such as deprecation rules should be ignored. I think I can easily add that myself after this PR is merged.

pavel-kirienko:
I am not sure if it is safe to rely on regular expressions for specification purposes given that the regexp syntax is not explicitly defined. Perhaps we could just put an explicit reference to IEEE POSIX ERE? https://en.wikipedia.org/wiki/Regular_expression#Standards http://pubs.opengroup.org/onlinepubs/000095399/basedefs/xbd_chap09.html

specification/dsdl/dsdl.tex
```diff
@@ -132,30 +132,100 @@ \section{Syntax}\label{sec:dsdl_syntax}
...
+\texttt{bool} & Primitive type. \\
+\texttt{true} & \texttt{bool} variant. \\
+\texttt{false} & \texttt{bool} variant. \\
+\texttt{con} & Reserved. \\
+\texttt{prn} & Reserved. \\
+\texttt{aux} & Reserved. \\
+\texttt{nul} & Reserved. \\
+\texttt{com\textbf{X}} & Reserved. \\
+\texttt{lpt\textbf{X}} & Reserved. \\
+\end{UAVCANSimpleTable}
```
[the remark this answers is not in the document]
kjetilkjeka: Yes, let's reserve these. I agree.

> The word offset should be un-reserved because it is too useful on its own (e.g. see the file transfer types). As it is only useful in expressions, it is perfectly safe to use this word for naming attributes and types.

Then we would have a contextual keyword, which makes implementation unnecessary hard/messy. I don't think this is a good idea, let's instead use a different word or a symbol for this then?

specification/dsdl/dsdl.tex
```diff
@@ -318,6 +392,30 @@ \subsubsection{Deprecated}
...
+if the value is \verb|false| the resulting type must be considered invalid, and steps must be taken to avoid usage.
+Both checking the assertions before compilation and compiling the assertions into compile time assertions in the
+resulting languages is considered legal way to handle this as long as it results in no way to use the resulting type.
+
+There exists a few magical keywords only available for assertions.
+This is a deliberate choice made to allow implementation of constant evaluation
+without needing the in-depth anal\begin{UAVCANSimpleTable}{Compatibility between associated const and const type}{|l l|}
+\textbf{Keyword} & \textbf{Description} \\
+\texttt{offset} & Evaluates into the set of possible bit offsets at the point of evaluation.
+\end{UAVCANSimpleTable}ysis required when checking assertions.
```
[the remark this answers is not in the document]
kjetilkjeka: Just started to use vscode, I really like it but weird things happen when I cut/copy/paste stuff, which I sometimes seems to be related to multi line cursors. I should probably figure it out before, sodomy ends up as a required part of the uavcan specification.

specification/dsdl/dsdl.tex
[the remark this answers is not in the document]
kjetilkjeka: I also added `enum`

kjetilkjeka:
> We should reflect all useful points from https://forum.uavcan.org/t/new-dsdl-directive-for-simple-compile-time-checks/190/ in this chapter.

I think the rest of this would require non-normative boxes (which i guess is ready now).

> I just realized that this PR seems to be focused exclusively on const and expressions (yes I just read the title). In which case my suggestion to introduce other stuff such as deprecation rules should be ignored. I think I can easily add that myself after this PR is merged.

Like you probably have seen there are a few days between i get to work on this specification so if you have the time I think it makes sense to merge this PR and for you to add the remaining?

kjetilkjeka:
> I am not sure if it is safe to rely on regular expressions for specification purposes given that the regexp syntax is not explicitly defined. Perhaps we could just put an explicit reference to IEEE POSIX ERE? https://en.wikipedia.org/wiki/Regular_expression#Standards http://pubs.opengroup.org/onlinepubs/000095399/basedefs/xbd_chap09.html

I don't think anything fancy is used in regards to regex, if that statement was to be found somewhere applying for the whole specification i think everything would be fine. Something like: "Wherever regex are used, they adhere to the IEEE POSIX ERE notation."

specification/dsdl/dsdl.tex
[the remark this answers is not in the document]
pavel-kirienko: > Then we would have a contextual keyword, which makes implementation unnecessary hard/messy.

Eh, I don't think it does. `offset` doesn't need to be a keyword, it is an identifier which refers to a particular value. Like a field. It doesn't require any special handling logic, and hence it doesn't need to be reserved. Replacing `offset` with a special character or prefixing it with special characters is probably not a good idea because readability will suffer.

pavel-kirienko:
> Like you probably have seen there are a few days between i get to work on this specification so if you have the time I think it makes sense to merge this PR and for you to add the remaining?

Will merge as soon as we resolve the `offset` issue.

specification/dsdl/dsdl.tex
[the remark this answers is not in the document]
pavel-kirienko: > If offsets are replaced with constant set literals early on, it needs to be figured out whether we're in an assertion context before they can be replaced or not.

I don't think we gain much by allowing early replacement. Our grammar remains simple enough even if we postpone offset handling until the point where assertions are checked. Assertion checking must be postponed until the very end of the parsing process anyway, because until all of the fields are processed, their sizes and offsets cannot be known.

Parsers can treat `@assert` (and `@print`) directives as black boxes until everything else is parsed, not even looking at their contents until it can't be delayed anymore. This removes the need for special handling of any keywords inside. Perhaps you have a different architecture in mind?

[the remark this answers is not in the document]
kjetilkjeka: In the name of keeping DSDL simple, we should use a keyword that is fine to reserve. Things will probably be fine with context dependent keywords, but it's not a place where one normally wants to start.

The worst thing I can think off in 30 seconds is the following ambigouty:
```
uint8 offset = 12
uint8 field
@assert offset == {8}
```

We have to do one of the following:
 - Disallow the situation by making offset a keyword
 - Define that these "magic words" take precedence over consts.

This was pretty much all i had. If you insist on removing `offset` as a keyword, just say so and I will do it.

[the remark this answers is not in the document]
pavel-kirienko: Ah, you have a point. When I originally proposed assertion checks, I intended them to be much simpler than what they are becoming now; particularly, I did not plan for the scenario you described to be possible. I realize though that we later discussed it further and agreed that a more advanced expression evaluation capability should be implemented, so my original outlook is no longer valid.

Okay, can we somehow keep the keyword sensible and at the same time avoid robbing definitions from useful words like `offset`? Ugly things come to mind like `_offset`. The screaming snake case does not seem to help because there can always be a constant named `OFFSET`, we don't want to prohibit that either. Perhaps we could pick a less useful keyword and use it as an umbrella identifier for referring to special entities constructed by the parser; e.g. instead of `offset` we would say `meta.offset`? Or just `.offset`, although it might be a bit opaque.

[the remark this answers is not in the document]
kjetilkjeka: > Perhaps we could pick a less useful keyword and use it as an umbrella identifier for referring to special entities constructed by the parser; e.g. instead of offset we would say meta.offset? Or just .offset, although it might be a bit opaque.

What about introducing a concept of "meta values", and use a specific symbol for them (similar to compiler directives like `@assert`).  We can then call `offset` a location dependent meta value. We can later introduce things as `MAJOR_VERSION`, `PI`, etc if we want to.

What about some sort of bracket notation `<offset>`? Or a dollar symbol `$offset`?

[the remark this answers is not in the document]
pavel-kirienko: So we are choosing between `meta.offset` and `$offset`, if we were to ignore the ugly options like `_offset` or `<offset>`. I generally dislike overly short constructs because they might undermine clarity, but in this case choosing dollars over an umbrella identifier I suggested earlier might be sensible because it works well in other contexts: the dollar allows us to un-reserve several keywords that are used as special-purpose identifiers: `root`, `super`, `self`, and use `$root`, `$super`, and `$self` instead.

So the list of additional reserved keywords will include: `enum`, `struct`, `const`, `template`, and the current offset will be referred to as `$offset`. Any objections?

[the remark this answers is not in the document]
kjetilkjeka: Great!

I would like to keep `root`, `super` and `self` for now. If only for being super likely to collide with the target language. What do you think?

specification/dsdl/dsdl.tex
thirtytwobits: C++ allows using not, or, and, etc instead of symbols (https://en.cppreference.com/w/cpp/language/operator_alternative) so this isn't just the pythonic way of doing things. I'm for words if it doesn't impeded RUST compatibility.

specification/dsdl/dsdl.tex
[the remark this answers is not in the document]
thirtytwobits: I'm so used to '$' meaning "variable" (perl, bash, and Make all use this syntax) that it would confuse the crap out of me to see it used as "meta keyword". I'd vote for one of the following:
```
.offset      # dot syntax as if this is a member of the dsdl global namespace and we are
             # descending one level of variable scope to access it.

::offset     # Similar to dot but using C++-like namespace syntax

.            # I proposed this before; to use ld syntax for "the currently computed memory location"

dsdl.offset  # explicit access to a global "dsdl" namespace

dsdl::offset # same as above but with C++-like syntax.
```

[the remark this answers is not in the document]
pavel-kirienko: > I would like to keep `root`, `super` and `self` for now. If only for being super likely to collide with the target language. What do you think?

Okay, why `root` though? Is there any mainstream language where this is a reserved keyword? The other two seem fine.

> I'm so used to '$' meaning "variable" (perl, bash, and Make all use this syntax) that it would confuse the crap out of me to see it used as "meta keyword".

Okay, so we're back to `meta.offset`, except that instead of `meta` there's now `dsdl`, which is fine. We could also consider using `type`, which has the advantage of being a reserved keyword/identifier in Rust/Python: `type.offset`.

So, `dsdl` or `type`?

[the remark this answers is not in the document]
thirtytwobits: I'm fine with either but I think I prefer dsdl.

specification/dsdl/dsdl.tex
pavel-kirienko: I also like words. Even if we don't reserve `not`, `or`, and friends, they still can't be safely used because they are likely to be reserved in the target language (C/C++/Python/Ada/Verilog).

specification/dsdl/dsdl.tex
```diff
@@ -132,30 +132,102 @@ \section{Syntax}\label{sec:dsdl_syntax}
...
+4 & \texttt{==}   & \texttt{const bool}   & \texttt{const bool}  & \texttt{const bool}    & Boolean equality \\
+4 & \texttt{!=}   & \texttt{const bool}   & \texttt{const bool}  & \texttt{const bool}    & Boolean inequality \\
+4 & \texttt{==}   & \texttt{const char}   & \texttt{const char}  & \texttt{const bool}    & Character equality \\
+4 & \texttt{!=}   & \texttt{const char}   & \texttt{const char}  & \texttt{const bool}    & Character inequality \\
+4 & \texttt{==}   & \texttt{\{T\}}        & \texttt{\{T\}}       & \texttt{const bool}    & Set equality \\
+4 & \texttt{!=}   & \texttt{\{T\}}        & \texttt{\{T\}}       & \texttt{const bool}    & Set inequality \\
+
+
+5 & \texttt{\&\&} & \texttt{const bool}   & \texttt{const bool}  & \texttt{const bool}    & Logical and \\
+5 & \texttt{\&\&} & \texttt{const bool}   & \texttt{const bool}  & \texttt{const bool}    & Logical or
```
pavel-kirienko: I suggest dropping the word `const` from the table because it is redundant. This is not executable code, so, fundamentally, all values we're dealing with are evaluated at the time of parsing, and so all values can be considered constant.

specification/dsdl/dsdl.tex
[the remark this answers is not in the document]
kjetilkjeka: `dsdl.type` is ambiguous with an associated const from a different file (that doesn't adhere to naming conventions) called `dsdl` (in the same root namespace). 

I would prefer not to start reserving file names as well. It's just one more concept that doesn't need to exist.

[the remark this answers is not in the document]
pavel-kirienko: > This is (I guess?) unless we enforce all dsdl definitions to be version, I don't remember what we ended up on.

Yes, currently we require that whenever you refer to a type, you specify its version explicitly, always. But even that doesn't matter; my understanding is that reserved keywords can't be used for naming types or namespaces, otherwise conflicts and confusion are bound to occur. Even if the DSDL spec allowed you to get away with naming your type, say, `struct`, it is unlikely that you will be able to generate code out of that without fixing your identifiers first, like `struct_` or `r#struct` or whatever suits your language. If we accept this reasoning, then it is safe to use `type.offset`.

> Remember the `offset` symbol is not a const or a variable. It is some sort of magic context dependent that the compiler have special handling rules for.

We could rename it for clarity, like `type.current_offset`? I don't like re-using `@` because this symbol is not really a directive (the semantics are quite different), and `@assert_offset` is unnecessarily resctictive -- there are likely to be other things to check in the future, not just offsets; besides, usage of that construct in arithmetic expressions would be impossible or weird.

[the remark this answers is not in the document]
pavel-kirienko: > It is some sort of magic context dependent that the compiler have special handling rules for.

If we were to emphasize the magic nature of this value, and if $ is not an option, we could go back to the other options suggested by Scott above. Of which I don't like `.` (dot) because of its syntactic ambiguity (can't be parsed using regular grammars), so we're left with `::offset` or even just plain `:offset`. Any objections to that?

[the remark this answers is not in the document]
thirtytwobits: I'm okay with either

pavel-kirienko:
Even if we haven't reached a definite conclusion, I think we're more or less on the same page, so I will take over from here. Await my PR soon.

specification/dsdl/dsdl.tex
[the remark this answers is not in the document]
kjetilkjeka: I'm also okay with either, small preference towards `:offset`

kjetilkjeka:
> Even if we haven't reached a definite conclusion, I think we're more or less on the same page, so I will take over from here. Await my PR soon.

Sorry to be unreachable, I was away for some days. Good that you were able to move on.

PR #36, opened 2018-12-02
Downgrade texlive to ubuntu 16.04 default to quickfix malformed PDFs

### LINKED ISSUE #35: Backport the LaTeX style fixes from zubaxdoc.cls
This is what the spec should be formatted like when built correctly:

image

This is what I'm getting after downloading build artifacts from GitLab:

image

Section headers are offset from the left for some reason, which breaks alignment and makes the doc look sloppy. I think it's important that we get consistent results on local machines and from the CI, otherwise, we might accidentally publish a bad doc.

kjetilkjeka:
Fixes #35


kjetilkjeka:
It looks like this PR fixes the problem for now.

We should look into upgrading to a newer version in the long run.

PR #38, opened 2018-12-05
Chapter 4 - transport layer

pavel-kirienko:


specification/transport_layer/can.tex
```diff
@@ -23,235 +23,252 @@ \subsection{CAN ID structure}
...
-\textbf{CAN ID Bit} &
-28 & 27 & 26 & 25 & 24 & 23 & 22 & 21 & 20 & 19 & 18 & 17 & 16 & 15 &
-14 & 13 & 12 & 11 & 10 &  9 &  8 &  7 &  6 &  5 &  4 &  3 &  2 &  1 &  0 \\
-\hline
-
-\end{tabular}
-}
-
-\caption{CAN ID structure}\label{fig:can_id_structure}
+    \centering
```
thirtytwobits: I'm super-nervous that, should this protocol see widespread use, the 128 node limit will vex us. I'll defer to your judgment here but my main concern arises from using UAVCAN on linux. SocketCAN creates a virtual CAN network which makes it easy to suddenly have hundreds of nodes if the system is designed like a ROS system where there are a multitude of small, interconnected functions. Using such an architecture and for a complex system it'd be easy to run out of node identifiers. Are we thinking that such a system is ill-advised or do we think that a NAT-like scheme might be needed to support this type of use case (i.e. where each linux network was isolated somehow by router nodes that could forward CAN messages across physical busses?

specification/transport_layer/can.tex
```diff
@@ -23,235 +23,252 @@ \subsection{CAN ID structure}
...
+            \\\hline
+
+            \textbf{CAN ID byte} &
+            \multicolumn{5}{c|}{3} & \multicolumn{8}{c|}{2} & \multicolumn{8}{c|}{1} & \multicolumn{8}{c|}{0}
+            \\\hline
+
+            \multicolumn{30}{c}{} \\ \hline % Table separator
+
+            %
+            % Anonymous message transfer
```
thirtytwobits: Should we combine the anonymous and normal message diagrams into one? (I find myself comparing each bit looking for what is different which is a little confusing) Basically we'd be saying there are two types; messages and service transfers, but that the message type comes in two flavors; regular and anonymous.

Just a thought. Not sure if this is a good idea or not.

specification/transport_layer/can.tex
```diff
@@ -23,235 +23,252 @@ \subsection{CAN ID structure}
...
+                                                      (or use according to a newer specification)
+                                                      if this field is not 1. \\
 \end{UAVCANSimpleTable}
 
 \subsubsection{Transfer priority}
 
 Valid values for priority range from 0 to 7, inclusively,
 where 0 corresponds to the highest priority, and 7 corresponds to the lowest priority.
-Mnemonics for transfer priority levels are provided in the section \ref{sec:transfer_prioritization}.
+Mnemonics for transfer priority levels are provided in the section \ref{sec:transfer_prioritization},
```
thirtytwobits: Isn't this section supposed to be non-normative? If so do we have a way of indicating this yet?

specification/transport_layer/can.tex
```diff
@@ -23,235 +23,252 @@ \subsection{CAN ID structure}
...
+    4 & Nominal \\
+    5 & Low \\
+    6 & Slow \\
+    7 & Optional \\
+\end{UAVCANSimpleTable}
+\end{minipage}
 
 In multi-frame transfers, the value of the priority field must be identical for all frames of the transfer.
 
 Shall there be multiple transfers of different types at the same priority level contesting for the bus access,
```
thirtytwobits: (grammar)

```suggestion
When multiple transfers of different types with the same priority contest for bus access
```

specification/transport_layer/can.tex
```diff
@@ -23,235 +23,252 @@ \subsection{CAN ID structure}
...
+transfers with lower port ID values preempt those with higher port ID values.
 
 \subsubsection{Payload discriminator}\label{sec:can_payload_discriminator}
 
 CAN bus does not allow different nodes to transmit CAN frames with different data field values under the same CAN ID.
 Owing to the fact that the CAN ID includes the node ID value of the transmitting node,
 this restriction does not affect regular UAVCAN transfers.
 However, anonymous message transfers would violate this restriction,
-because they all share the same node ID of zero.
+because they don't have unique node ID.
```
thirtytwobits:
```suggestion
because they don't have a unique node ID.
```

specification/transport_layer/can.tex
```diff
@@ -23,235 +23,252 @@ \subsection{CAN ID structure}
...
-Node ID is represented by a 7-bit unsigned integer value; zero is reserved.
-A node ID of zero is used to represent either an unknown node or all nodes, depending on the context.
-
-As such, for anonymous message transfers, the source node ID field is always set to zero.
-By observing a source node ID of zero, receiving nodes can distinguish
-anonymous message transfers from other types of transfers.
+transfers are prioritized according to the port ID\footnote{Subject ID or service ID.}:
+transfers with lower port ID values preempt those with higher port ID values.
 
 \subsubsection{Payload discriminator}\label{sec:can_payload_discriminator}
```
thirtytwobits: We should have non-normative guidance about the use of anonymous messages somewhere before this section. Namely we should point out that there is only one known reason to use them in the current specification (i.e. the dynamic node protocol) and that they are not generally useful. The problem is we have this big section on discriminators which, by the weight of it's complexity, might suggest that this message type is actually very common. We should dispel any such notion. It would be a real tragedy if someone designed an extensive system using all anonymous nodes (perhaps to get around the 128 node limit?) and then complained about the dismal performance.

specification/transport_layer/can.tex
```diff
@@ -261,23 +278,28 @@ \subsubsection{Payload discriminator}\label{sec:can_payload_discriminator}
...
+Since the discriminator is only seven bits long (128 possible values),
+a collision where multiple nodes emit CAN frames with different data but same CAN ID is likely to happen
+despite the randomization measures described earlier.
 Therefore, the protocol must account for possible errors on the CAN bus triggered by CAN ID collisions.
 In order to comply with this requirement,
 UAVCAN requires all nodes to immediately abort transmission of all anonymous transfers once an error on
 the CAN bus is detected.
 This measure allows the protocol to prevent the bus deadlock that may occur if the automatic
 retransmission on bus error is not suppressed.
+Applications should use some form of CSMA/CD\footnote{Carrier-sense multiple access with collision detection
```
thirtytwobits: This is a very confusing statement to encounter given that CAN is a CSMA protocol. We should acknowledge that this is an application-level implementation of CSMA on top of a bus that does CSMA in hardware.

specification/transport_layer/can.tex
pavel-kirienko: Here's a counter-thought: should we split request and response for extra clarity instead?

specification/transport_layer/can.tex
pavel-kirienko: Yes, it's supposed to be non-normative. We don't have the right formatting primitives in place yet, we'll fix that later. Right now we should be focused on the content.

specification/transport_layer/can.tex
thirtytwobits: Seems reasonable.

specification/transport_layer/can.tex
thirtytwobits: I'd rather see the message types combined but if not then your suggestion is good as well.

kjetilkjeka: :+1: for the status quo.

I think anonymous is different enough for splitting and request/response is similar enough to keep together. 

I also think that we're down to a question of preference and any of the suggestions will make do.

specification/transport_layer/can.tex
```diff
@@ -261,23 +278,28 @@ \subsubsection{Payload discriminator}\label{sec:can_payload_discriminator}
...
 
-Since the discriminator is only 10 bits long,
-the probability of having multiple nodes emitting CAN frames with the same CAN ID
-but different data can exceed 0.1\%, which is significant.
+Since the discriminator is only seven bits long (128 possible values),
+a collision where multiple nodes emit CAN frames with different data but same CAN ID is likely to happen
+despite the randomization measures described earlier.
 Therefore, the protocol must account for possible errors on the CAN bus triggered by CAN ID collisions.
 In order to comply with this requirement,
 UAVCAN requires all nodes to immediately abort transmission of all anonymous transfers once an error on
```
kjetilkjeka: Maybe include "If the transport protocol use automatic retransmission, this must be disabled for anonymous frames"

specification/transport_layer/can.tex
kjetilkjeka: Is it better to define a few rules?
 - An anonymous message that fails to be transmitted must not be retransmitted with the same discriminator unless a scheme for an indeterministic (guaranteed low correlation with other nodes on the bus) chosen to wait is used.
 - An anonymous message that fails to be transmitted must not be retransmitted immediately unless a scheme for indeterministic (guaranteed low correlation with other nodes on the bus) discriminator selection is chosen.

specification/transport_layer/can.tex
pavel-kirienko: I think it would be a non-issue if we went back to the portrait CAN ID diagram we used to have before https://github.com/UAVCAN/specification/pull/31. Kjetil, are you still vehemently opposed to using portrait diagrams?

kjetilkjeka: I just find the horizontal diagrams much more readable. I honestly think by listing the bits vertically it would take me 2-3 times as long to wrap my head around it.

I think both 2, 3 or 4 horizontal figures would be more readable.

pavel-kirienko: Okay, I have collapsed this stuff into just two diagrams as Scott suggested. I found this to be even more sensible if we renamed "payload discriminator" into "source node ID stub", which I just did. Please review.

specification/transport_layer/can.tex
pavel-kirienko: It seems sufficient to say that we need CSMA/CD; spelling out the implementation details seems unnecessary.

specification/transport_layer/can.tex
pavel-kirienko: Added.

pavel-kirienko:
Merged master into this branch to fix the CI.

specification/transport_layer/can.tex
thirtytwobits: 👍 but (because I like green bike sheds) a "stub" is a remnant or abbreviation of a full thing. The anonymous identifier is neither since the sender has not node ID. Perhaps calling it the "source node dummy ID" or "source node psedo-ID" would be more accurate?

specification/transport_layer/can.tex
thirtytwobits: Saw the footnote. 👍

specification/transport_layer/can.tex
pavel-kirienko: Replaced with pseudo-ID.

specification/transport_layer/can.tex
pavel-kirienko: Moved CSMA/CD to the footnote.

pavel-kirienko:
All existing comments have been addressed it seems (except for the non-normative section formatting), so I am merging this now. If you have additional comments, please open tickets or send PR.

PR #39, opened 2018-12-11
Chapter 6 - physical layer

pavel-kirienko:
The changes are extremely trivial, at least so far.

### DIFF
--- a/specification/physical_layer/can.tex
+++ b/specification/physical_layer/can.tex
@@ -15,8 +15,8 @@ for the CAN bus transport implementation.
 Other connector types may be added in future revisions of the specification.
 
 It is highly recommended to provide two identical parallel connectors for each CAN interface per device,
-so that the device can be connected to the bus without the need to use \mbox{T-connectors}.
-\mbox{T-connectors} should be avoided when possible because generally they add an extra point of failure,
+and not using \mbox{T-connectors}.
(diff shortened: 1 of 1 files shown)


specification/physical_layer/can.tex
```diff
@@ -15,8 +15,8 @@ \subsection{Physical connector specification}
 Other connector types may be added in future revisions of the specification.
 
 It is highly recommended to provide two identical parallel connectors for each CAN interface per device,
-so that the device can be connected to the bus without the need to use T-connectors.
-T-connectors should be avoided when possible because generally they add an extra point of failure,
+so that the device can be connected to the bus without the need to use \mbox{T-connectors}.
```
thirtytwobits:
```suggestion
and not using \mbox{T-connectors}.
```

specification/physical_layer/can.tex
```diff
@@ -15,8 +15,8 @@ \subsection{Physical connector specification}
...
+\mbox{T-connectors} should be avoided when possible because generally they add an extra point of failure,
```
thirtytwobits:
```suggestion
\mbox{T-connectors} should be avoided because they add another point of failure,
```

pavel-kirienko:
@kjetilkjeka I suppose the changes here are trivial enough for this to be safe to merge. If you find any issues, please open a ticket.

PR #44, opened 2018-12-27
Application layer (chapter 5) and List of standard data types (chapter 6)

pavel-kirienko:


render_dsdl.py
```diff
@@ -0,0 +1,273 @@
...
+    return '\n'.join([
+        r'\begin{minted}[%s]{python}' % minted_params,
+        open(t.source_file_path).read(),
+        r'\end{minted}',
+    ])
+
+
+try:
+    index_only = False
+    if '--index-only' in sys.argv:
```
thirtytwobits: why not use argparse?

.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "dsdl"]
```
thirtytwobits: I'm a little conflicted about this. Conceptually dsdl depends on a give version of the specification which makes this, technically, a circular dependency. I'm also not sure that we want to treat the standard, regulated type documentation so differently. i.e. we might want to provide your render-to-latex script as something any dsdl repo can use to generate documentation.

specification/application_layer/conventions.tex
```diff
@@ -0,0 +1,238 @@
+\section{Application-level conventions}\label{sec:application_level_conventions}
```
thirtytwobits: This all seems pretty close to what we had before and what we've discussed in the forums. Mostly just a reminder here that this needs careful copy-editing to simplify and properly voice all the English.

specification/standard_data_types/standard_data_types.tex
```diff
@@ -0,0 +1,20 @@
+\chapter{List of standard data types}
+
+This chapter contains the full list of standard data types defined by the UAVCAN specification.
+The source text of the DSDL data type definitions provided here is also available via the
+official project website at \href{http://uavcan.org}{uavcan.org}.
+
+Regulated non-standard definitions\footnote{I.e., public definitions contributed by vendors and other users
+of the specification, as explained in section \ref{sec:basic_concepts_data_type_regulation}.}
+are intentionally not included in this list, as they are of limited importance for the specification.
```
thirtytwobits: We should find a nicer way to phrase this. While semantically true, it's not polite to tell vendors that their carefully designed datatypes are of limited importance.

render_dsdl.py
```diff
@@ -0,0 +1,273 @@
...
+
+# We avoid using longtabu unless we really have to because it's a bloody disaster. Its caption is always misaligned
+# vertically (requires hacking with \abovecaptionskip to fix) and it tends to run over text and footnotes below it.
+table_environment = 'tabu' if is_nested_namespace else 'longtabu'
+
+if is_nested_namespace:
+    # Confine nested namespace indexes to one page, assuming that they are always compact enough to fit.
+    print(r'{\parindent=-\leftskip\begin{minipage}{\textwidth}\centering')
+
+print(r'\begin{ThreePartTable}')
```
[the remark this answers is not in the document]
pavel-kirienko: Sure. I would avoid jinja though because it's an external dependency, and instead used something like [Pyratemp](https://www.simple-is-better.org/template/pyratemp.html) because it's MIT-licensed and distributed as a one-file module which can be re-distributed together with the script (it's used in libuavcan and libcanard for code generation). That said, I don't want to change the script now.

specification/application_layer/application_layer.tex
```diff
@@ -1,5 +1,23 @@
 \chapter{Application layer}\label{sec:application_layer}
 
-\section{Application-level conventions}\label{sec:application_level_conventions}
+Previous chapters of this specification define a set of basic concepts that are the foundation of the protocol:
```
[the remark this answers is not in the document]
pavel-kirienko: I should pass this comment over to @hamishwillee

specification/application_layer/application_layer.tex
```diff
@@ -1,5 +1,23 @@
...
+Unless stated otherwise, all of the high-level functions described in this chapter are optional,
+meaning that they do not necessarily have to be supported or implemented for an application or system to be
+UAVCAN-compliant.
+Excepting a very few mandatory functions and rules, users of UAVCAN are allowed to pick only those
+that are needed and suitable for their applications.
+
+If a designer chose to deviate from the rules and conventions presented in this chapter,
+every such deviation must be explicitly documented.
+
+\clearpage\input{application_layer/conventions.tex}
```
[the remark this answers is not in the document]
pavel-kirienko: No point spilling floats into the next section, so `\clearpage` should be used here instead of `\newpage`.

render_dsdl.py
pavel-kirienko: ¯\\\_(ツ)\_/¯

specification/application_layer/conventions.tex
pavel-kirienko: Noted.

render_dsdl.py
pavel-kirienko: I should probably clarify here that I preferred adding three lines of code to handle the argument manually over adding six lines of code to do that using the library. KISS

a1ien:
Adding dsdl broken gitlab build. Job not produce artifact https://gitlab.com/Uavcan/specification/-/jobs/141696034

pavel-kirienko:
Thanks @a1ien, fixed.

PR #46, opened 2019-04-08
Chapter 3 - DSDL specification

pavel-kirienko:
At a first glance it seems to be complete content-wise and the structure is probably the best I could muster without rewriting the whole thing from scratch again. I am going to re-read it again tomorrow and fix the remaining inconsistencies (especially inconsistent term usages) and missed details.

The most important task right now is to ensure that the DSDL specification is complete and all relevant aspects of the language are described unambiguously. I would like you to review the chapter and try and answer the following questions:

* Are all concepts of the language described clearly?
* Are there any terms or definitions that are used consistently? Please disregard their usage in other chapters for now.
* Are there any ambiguous or contradictory requirements or definitions?
* Is it possible to simplify the specification or reduce its verbosity by introducing new or more specific definitions?
* Can the number of forward references be substantially reduced by rearranging the sections?
* Are there sections that should be extended with examples?
* Is there any non-normative text that is not enclosed in shaded blue remark boxes or extracted into footnotes?

If you find bad wording, misspelled words, bad grammar, missing articles (sorry), and similar language-level errors, please ignore them for now unless they are also related to any of the above issues.

Don't look at the diff, it's huge and mostly meaningless; instead, use this prebuilt PDF: 
[UAVCAN_Specification.pdf](https://github.com/UAVCAN/specification/files/3055966/UAVCAN_Specification.pdf)

After this PR is finished and merged, I am going to introduce a few sweeping changes in the document to fix inconsistent term usage and some formatting issues. The introductory chapter will be expanded with a very brief section on motivation (one or two sentences), and the section listing the document conventions also requires extension (for example, the DSDL chapter uses notation like **Container<sub>\<Element\></sub>** for higher-order type parametrization, but it is not explicitly introduced anywhere yet). Also, I am hoping to bring the total number of pages under 100 (many a reader would be repelled by the three-digit number of pages) by changing the page geometry somewhat.

@kjetilkjeka @thirtytwobits please review.


specification/dsdl/architecture.tex
```diff
@@ -0,0 +1,233 @@
...
+one of them applies to the request object (transferred from client to server),
+and the other applies to the response object (transferred from the server back to the client).
+
+Following the deterministic nature of UAVCAN, the size of a serialized representation of any
+message or service object is bounded within statically known limits.
+Variable-size entities always have a fixed size limit defined by the data type designer.
+
+DSDL definitions are strongly statically typed.
+
+DSDL provides well-defined means of data type versioning, which enable data type maintainers to introduce changes
```
thirtytwobits:
```suggestion
DSDL provides a well-defined means of data type versioning, which enables data type maintainers to introduce changes
```

```suggestion
DSDL provides data type versioning allowing type maintainers to introduce changes
```

specification/dsdl/architecture.tex
```diff
@@ -0,0 +1,233 @@
...
+    $$
+    \caption{Data type definition file name structure.\label{fig:dsdl_definition_file_name_structure}}
+\end{figure}
+
+DSDL namespaces are represented as directories, where one directory defines exactly one namespace, possibly nested.
+The name of the directory defines the name of its data type name component.
+A directory defining a namespace will always define said namespace in its entirety,
+meaning that the contents of a namespace cannot be spread across different directories sharing the same name.
+One directory cannot define more than one level of
+nesting\footnote{For example, ``\texttt{foo.bar}'' is not a valid directory name.
```
thirtytwobits: But most operating systems allow this as a folder name. Should we define a standard transformation from valid posix path names into valid dsdl that can be performed automatically as a pre-processing step (i.e. before we put the type into the grammar parser we perform a platform to internal representation transform). So "foo.bar" would become something like "foo_bar" etc.

specification/dsdl/architecture.tex
```diff
@@ -0,0 +1,233 @@
...
+Data type definitions can also contain various auxiliary elements reviewed later,
+such as deprecation markers (notifying its users that the data type is no longer recommended for new designs)
+or assertions (special statements introduced by data type designers
+which are statically validated by DSDL processing tools).
+
+\subsection{Serialization}
+
+Every object that can be exchanged between UAVCAN nodes has a well-defined \emph{serialized representation}.
+The value and meaning of an object can be unambiguously recovered from its serialized representation,
+provided that the type of the object is known.
```
thirtytwobits: We need a section describing how types are "known". This is especially significant now that we no longer have "message ids" but are using port ID instead. With this we've weakened the link between a CAN identifier and a DSDL type making it possible, for example, to exchange the same type over different ports. We've talked about how such mappings will have to be installed into a system but I'm not sure we've discussed this in the document? If we have, sorry, just add cross-reference to this discussion.

specification/dsdl/grammar.parsimonious
```diff
@@ -0,0 +1,178 @@
...
+
+statement_attribute = statement_constant
+                    / statement_field
+                    / statement_padding_field
+
+statement_constant      = type _ identifier _? "=" _? expression
+statement_field         = type _ identifier
+statement_padding_field = type_void ""  # The trailing empty symbol is to prevent the node from being optimized away.
+
+statement_service_response_marker = "---"  # Separates request/response, specifies that the definition is a service.
```
thirtytwobits: I'd make this "3 or more" dashes. Style may dictate 3 but I don't see a benefit in being pedantic in the grammar about this.

specification/dsdl/grammar.parsimonious
```diff
@@ -0,0 +1,178 @@
...
+
+statement_directive = statement_directive_with_expression
+                    / statement_directive_without_expression
+statement_directive_with_expression    = "@" identifier _ expression  # The expression type must match the directive.
+statement_directive_without_expression = "@" identifier
+
+identifier = ~r"[a-zA-Z_][a-zA-Z0-9_]*"
```
thirtytwobits: Seems like this belongs up at the top since it's reused in several of your sections below?

specification/dsdl/grammar.tex
```diff
@@ -0,0 +1,253 @@
...
+
+A boolean literal is denoted by the keyword ``\verb|true|'' or ``\verb|false|''
+represented by an instance of primitive type ``\verb|bool|'' (section \ref{sec:dsdl_primitive_types})
+with an appropriate value.
+
+\subsubsection{Numeric literals}
+
+Integer and real literals are represented as instances of type ``\verb|rational|'' (section \ref{sec:dsdl_rational}).
+
+The digit separator character ``\verb|_|'' (underscore) does not affect the interpretation of numeric literals.
```
thirtytwobits: underscore to separate digits? Don't you mean comma?

specification/dsdl/expression_types.tex
```diff
@@ -0,0 +1,223 @@
...
+Results of such expressions can be used to define various constant properties,
+such as array length boundaries or values of constant attributes.
+
+Expression types are specified in this section.
+Each expression type has a formal DSDL name for completeness;
+even if such types can't be used to define attributes,
+a well-defined formal name allows DSDL processing tools to emit well-formed
+and understandable diagnostic messages.
+
+\subsection{Rational number}\label{sec:dsdl_rational}
```
thirtytwobits: see https://forum.uavcan.org/t/rational-expressions-in-dsdl-constants/444/10

specification/dsdl/expression_types.tex
```diff
@@ -0,0 +1,223 @@
...
+
+At the time of DSDL definition processing, integer and real numbers are represented internally as rational numbers
+where the range of numerator and denominator is unlimited\footnote{%
+Technically, the range may only be limited by the memory resources available to the DSDL processing tool.
+DSDL processing tools are not permitted to introduce any implicit rational number transformations that
```
thirtytwobits: This is a requirement of the specification  and should be pulled out of a footnote and placed into a normative section in the text.

specification/dsdl/serializable_types.tex
```diff
@@ -0,0 +1,285 @@
...
+instances of which can only exist while DSDL definitions are being evaluated.
+The data serialization rules are defined in the section \ref{sec:dsdl_data_serialization}.
+
+\subsection{Void types}
+
+Void types are used for padding purposes.
+As will be explained in later sections, it is desirable to align field attributes at byte boundaries;
+void types can be used to facilitate that.
+
+Void-typed field attributes are set to zero when an object is serialized and ignored when it is deserialized.
```
thirtytwobits: Do we have to require that they are zeroed? Can't we just specify that the actual value of the void types is undefined?

specification/dsdl/serializable_types.tex
```diff
@@ -0,0 +1,285 @@
...
+Void types can be used to reserve space in data type definitions for possible use in later versions of the data type.
+
+The DSDL name pattern for void types is as follows: ``\verb|void[1-9]\d*|'',
+where the trailing integer represents its width, in bits,
+ranging from 1 to 64, inclusive.
+
+Void types can be referred to directly by their name from any namespace.
```
thirtytwobits: Huh. I thought void types were anonymous. We can actually read these values?

specification/dsdl/serializable_types.tex
```diff
@@ -0,0 +1,285 @@
...
+\immediate\input{../latex.tmp}
+
+\subsection{Array types}
+
+An array type represents an ordered collection of values.
+All values in the collection share the same type, which is referred to as \emph{array element type}.
+The array element type can be any type except:
+\begin{itemize}
+    \item void type;
+    \item array type\footnote{Meaning that nested arrays are not allowed;
```
thirtytwobits: Huh. Why not?

specification/dsdl/serializable_types.tex
```diff
@@ -0,0 +1,285 @@
...
+For example, if a data type version 0.1 is of a message kind, all later versions of it must be messages, too.}.
+
+A message data type defines one data schema.
+
+A service data type contains two data schema definitions:
+one for service request object, and one for service response object, in that order.
+The two schemas are separated by the service response marker (``\verb|---|'') on a separate line.
+
+The presence of the service response marker indicates that the data type definition at hand is of the service kind.
+There may be no more than one service response marker in a given definition.
```
thirtytwobits:
```suggestion
Only one service response marker shall appear in a given definition.
```

specification/dsdl/serializable_types.tex
```diff
@@ -0,0 +1,285 @@
...
+\footnote{For example, from user-provided configuration options.}
+the list of directories that are the roots of namespaces containing the referred dependencies.
+
+\begin{figure}[H]
+    $$
+    \underbrace{\texttt{\huge{uavcan.node.Heartbeat}}}_{\text{full data type name}}%
+    \texttt{\huge{.}}%
+    \underbrace{\texttt{\huge{1.0}}}_{\substack{\text{version} \\ \text{numbers}}}%
+    $$
+    \caption{Reference to an external composite data type definition.\label{fig:dsdl_nested_reference}}
```
thirtytwobits: Should we support a >=< syntax to allow some flexibility here? We might otherwise cause cascading changes that were unnecessary.

specification/dsdl/serializable_types.tex
```diff
@@ -0,0 +1,285 @@
...
+This tainting behavior is designed to prevent unexpected breakage of
+type hierarchies when one of the deprecated dependencies reaches its end of life.}.
+
+\subsubsection{Unions}\label{sec:dsdl_composite_tagged_unions}
+
+Any data schema definition can be supplied with a special directive (section \ref{sec:dsdl_directives})
+indicating that it forms a tagged union.
+
+A tagged union schema must contain at least two field attributes.
+A tagged union must not contain padding field attributes.
```
thirtytwobits:
```suggestion
A tagged union shall not contain padding field attributes.
```

specification/dsdl/serializable_types.tex
```diff
@@ -0,0 +1,285 @@
...
+the referring definition is malformed\footnote{%
+This tainting behavior is designed to prevent unexpected breakage of
+type hierarchies when one of the deprecated dependencies reaches its end of life.}.
+
+\subsubsection{Unions}\label{sec:dsdl_composite_tagged_unions}
+
+Any data schema definition can be supplied with a special directive (section \ref{sec:dsdl_directives})
+indicating that it forms a tagged union.
+
+A tagged union schema must contain at least two field attributes.
```
thirtytwobits:
```suggestion
A tagged union schema shall contain at least two field attributes.
```

specification/dsdl/serializable_types.tex
```diff
@@ -0,0 +1,285 @@
...
+A tagged union must not contain padding field attributes.
+
+The value of a tagged union object is a function of the field attribute which value it is currently holding
```
thirtytwobits: This sentence is missing commas but should be revised per my previous  comments about  tagged union  value functions.

specification/dsdl/versioning.tex
```diff
@@ -0,0 +1,411 @@
...
+
+$A$ and $B$ are said to be \emph{mutually bit-compatible} if
+their sets of serialized representations are equal.
+
+A \emph{variable-length} data type or schema is a serializable data type or schema
+whose set of serialized representations contains bit sequences of different lengths.
+Conversely, any data type or schema that is not variable-length is \emph{fixed-length}.
+
+\begin{remark}[breakable]
+    The following two definitions are bit-compatible:
```
thirtytwobits:
```suggestion
    The following two definitions are mutually bit-compatible:
```

specification/dsdl/versioning.tex
```diff
@@ -0,0 +1,411 @@
...
+    \item Comment changes of the DSDL source text are allowed as long as such changes
+    do not affect semantic compatibility of the definition.
+
+    \item A deprecation marker directive (section~\ref{sec:dsdl_directives}) can be added or removed\footnote{%
+    Removal is useful when a decision to deprecate a data type definition is withdrawn.}.
+\end{itemize}
+Addition or removal of the fixed port identifier is not permitted after a data type definition
+of a particular version is released.
+
+Therefore, substantial modifications of released data types are only possible by releasing
```
thirtytwobits: It's really hard to figure out what's being proposed the way it's worded in this section. We say "...the standard
requires that once a data type definition is released, its DSDL source text, name, version numbers, fixed
port ID, and other properties cannot undergo any modifications whatsoever" And then we say "...substantial modifications... are ...possible". These two statements contradict. I think we mean, in the previous paragraph, that you can only change a released type by releasing a new version of the type, correct?

specification/dsdl/conventions.tex
```diff
@@ -0,0 +1,115 @@
...
+Every attribute of the data type definition, and especially every field attribute of it,
+should have an associated comment explaining the purpose of the attribute, its semantics, usage patterns,
+assumptions, constraints, and any other pertinent information.
+Exception applies to attributes supplied with sufficiently descriptive and unambiguous names.
+
+% Field comment placement https://forum.uavcan.org/t/dsdl-documentation-comments/407
+
+\subsection{Optional value representation}
+
+Data structures may include optional field attributes that are not necessarily always populated.
```
thirtytwobits:
```suggestion
Data structures may include optional field attributes that are not always populated.
```
```suggestion
Data structures may include optional field attributes that are not necessarily populated.
```

These suggestions aside; I think we mean to say something like "There is no built-in support for optional field attributes but here's how you can fake it." If so then the former is normative and the latter is non-normative.

specification/dsdl/conventions.tex
```diff
@@ -0,0 +1,115 @@
...
+Alternatively, such one-element variable-length arrays can be replaced with two-field unions,
+where the first field is empty and the second field contains the desired optional value.
+The described layout is bit-compatible and semantically compatible with the one-element array described above,
+provided that the field attributes are not swapped.
+
+Floating-point-typed field attributes may be assigned the value of not-a-number (NaN) per IEEE 754
+to indicate that the value is not specified;
+however, this pattern is discouraged because some nodes may not support IEEE 754 NaN values internally,
+the value would still have to be transferred over the bus even if not populated,
+and special case values undermine type safety.
```
thirtytwobits: I've seen real defects in the field caused by mixed NaN support. Should we take a harder stance in the spec and require all nodes support it or that it is not supported/allowed?

specification/dsdl/conventions.tex
```diff
@@ -0,0 +1,115 @@
...
+        void7                           # It is recommended to ensure byte alignment.
+        MaybeMyType optional_field
+    \end{minted}
+
+    The shown approaches are mutually bit-compatible and semantically compatible.
+\end{remark}
+
+\subsection{Bit flag representation}
+
+The recommended approach to defining a set of bit flags is to dedicate a \verb|bool|-typed field attribute for each.
```
thirtytwobits: Similar comment here as above: I think we have a normative specification that "bit flags are not, intrinsically, supported" and a non-normative section about how they can be emulated.

specification/dsdl/conventions.tex
```diff
@@ -0,0 +1,115 @@
+\section{Conventions and recommendations}
+
+This section is dedicated to conventions and recommendations
+intended to help data type designers maintain a consistent style across the ecosystem
+and avoid some common pitfalls.
+Some of the conventions are mandatory to follow, others are optional.
```
thirtytwobits: I've added some comments below but generally I don't like mixing sparse bits of normative specification into larger blocks of non-normative. The former tends to get lost. Can we refactor this to have a short normative/mandatory section followed by the longer non-normative/optional sections?

specification/dsdl/versioning.tex
```diff
@@ -0,0 +1,411 @@
...
+        bool flag_psu_malfunction
+
+        void3   # Reserved for other flags
+
+        # Status flags (this is the recommended practice)
+        bool flag_cooling_system_a_active
+        bool flag_cooling_system_b_active
+    \end{minted}
+
+    Nodes using v1.0, v1.1, and v2.0 definitions can still coexist on the same network,
```
thirtytwobits: Please describe how the nodes determine what the major version is of a given message and how they must use this to lookup the correct response type major version.

specification/dsdl/attributes.tex
```diff
@@ -0,0 +1,193 @@
...
+        void6
+        # One of the main usages for _offset_ is statically proving that the following field is byte-aligned
+        # for all possible valid serialized representations of the preceding fields. It is done by computing
+        # a remainder as shown below. If the field is aligned, the remainder set will equal {0}. If the
+        # remainder set contains other elements, the field may be misaligned under some circumstances.
+        # If the remainder set does not contain zero, the field is never aligned.
+        @assert _offset_ % 8 == {0}
+        uint8 well_aligned   # Proven to be byte-aligned.
+    \end{minted}
+\end{remark}
```
thirtytwobits: __offset__ is sort of complex. Won't most authors really care about the maximum size anyway? Should we provide __max_offset__ that is a simple scalar value?

specification/dsdl/grammar.tex
pavel-kirienko: `123_456_789` is a literal number which matches `literal_integer_decimal` and is interpreted as 123456789. They use underscore in Python like this, too. In C++ they use the single quote character `'` instead, which might read better but it's incompatible with our usage of single quotes for string literals. Okay, it's compatible, but the grammar gets kind of weird.

`123,456,789` is an expression which matches `expression_list = (expression (_? "," _? expression)*)?`, not a numeric literal.

specification/dsdl/serializable_types.tex
pavel-kirienko: We shouldn't lift this requirement because it simplifies the extension of data types in a backward-compatible way by adding new fields in place of padding fields. Example:

```python
# Version 1.1
float64 a
void64
```

```python
# Version 1.2
float64 a
float32 b  # Messages v1.1 will be interpreted such that b = 0.0
void32
```

specification/dsdl/serializable_types.tex
pavel-kirienko: This is about types, not their values. This merely means that void types are not contained within some particular namespace, they are built-in, like `void` in C++.

specification/dsdl/serializable_types.tex
pavel-kirienko: We may allow them someday later. There are two weak reasons against nested arrays:

- The grammar gets slightly more complex. A naive addition of nested arrays would require us to make the array matching rule `type_array` left-recursive, which some parsers/generators may not support.  We can deal with this, though, if necessary, so it's not a huge deal.
- Variable-length nested arrays would often lead to misaligned data due to the length prefix being of weird bit width. Misaligned data is bad: https://github.com/UAVCAN/pydsdl/pull/24.

If one needs nested arrays, the following should be done instead:

```python
Nested.1.0[<=100] two_dimensional_array
```

```python
# Nested.1.0
void1  # Alignment
float16[<=100] one_dimensional_array
@assert _offset_.min % 8 == 0  # Yup, the data is aligned
```

Do you have any particular use cases in mind where nested arrays might be very desirable?

specification/dsdl/serializable_types.tex
pavel-kirienko: Replaced this with "At most one service response marker shall appear in a given definition."

specification/dsdl/serializable_types.tex
pavel-kirienko: Would be a mistake. See this and the next two posts: https://forum.uavcan.org/t/the-case-for-code-compatibility/194/10?u=pavel.kirienko

Currently, we provide a hard guarantee that once a given version of a data type is released, it is to be never changed again excepting whitespaces, comments, and `@deprecated`. It is important for maintainability and compatibility. Automatic switch to a newer version would remove this guarantee.

specification/dsdl/conventions.tex
pavel-kirienko: I'm worried that might be an overreach for this specification. Instead, I suggest we add a blue box right under this last sentence briefly explaining the dangers of mixed NaN interpretation. Do you have any particular wording in mind?

specification/dsdl/conventions.tex
pavel-kirienko: > I think we have a normative specification that "bit flags are not, intrinsically, supported"

But they are supported, are they not? We have `bool` which is encoded as a single bit. Should it say instead "Bit flags may be represented as fields of type `bool`" in the normative section? I don't see any issues with the current wording but I'm open to changing it.

specification/dsdl/attributes.tex
pavel-kirienko: As discussed, `_offset_` is crucial for static alignment analysis (and https://github.com/UAVCAN/pydsdl/pull/24, although it has no direct relation to the specification itself). The case of `_max_offset_` is also supported as `_offset_.max`.

specification/dsdl/architecture.tex
pavel-kirienko: I am not sure what would we gain by doing so? Implicit transformations would make the relationship between DSDL names and generated type names harder to follow. The objective here is to say that if one wants to define a data type `vendor.foo.bar.Baz` so that the namespaces `foo` and `bar` are otherwise empty, they would have to create a dedicated directory `foo/` and then `foo/bar/` instead of just `foo.bar/`. The latter form could be supported but it greatly complicates the mapping between namespaces and directories. Additionally, it would be unclear how to represent that in the generated code: using identical shortcuts, or by unrolling the directory tree into one-level-per-directory form?

specification/dsdl/architecture.tex
pavel-kirienko: This is very valid. Please review this commit I just pushed here: https://github.com/UAVCAN/specification/pull/46/commits/33b69a58ff9e0aff8117d404a143f9ea26042e09

specification/dsdl/versioning.tex
pavel-kirienko: Fixed here: https://github.com/UAVCAN/specification/pull/46/commits/b58830f1ee8614d29007daf88525faf7cd538116

The original wording seemed technically correct, but it could be hard to follow, indeed. It said "substantial modifications of released data types", but you seem to have read it as "substantial modifications of released data type definitions". A definition is virtually unchangeable, whereas a "data type" is a set of its definitions of different versions, as implied under "General assumptions".

specification/dsdl/versioning.tex
pavel-kirienko: I just added a reference to the relevant section in the Basic Concepts: https://github.com/UAVCAN/specification/pull/46/commits/c9ccffb145b167a92bc06df10bf503d94f0374e3

pavel-kirienko:
I seem to have either resolved or asked further questions on all of the raised issues except:

- https://github.com/UAVCAN/specification/pull/46#discussion_r277125331 (IEEE 754 constants)
- https://github.com/UAVCAN/specification/pull/46#discussion_r277125886 (mixed normative and non-normative sections)

I will be back to these around Wed/Thu; meanwhile, feel free to review my other changes.

specification/dsdl/architecture.tex
thirtytwobits: I'm not saying we support a shortcut where `vendor/foo.bar` defines a <strike>`vendor.foo.bar`</strike> namespace. I'm suggesting we may want to consider the frustration of a user where they have `google.com/wing` and expect the namespaces to be `"google.com".wing`. We could decide to simply say all `.` in a folder name become underscores. So wing would generate `google_com.wing` namespaces. Of course this is asymmetric so I can be convinced that it's just simpler to throw a very nicely worded error about why the dsdl generator refused to continue.

specification/dsdl/conventions.tex
thirtytwobits: If this is normative then we aren't recommending we are specifying. Which is it for bit flags?

specification/dsdl/conventions.tex
thirtytwobits: (Channeling a colleague of mine) shouldn't we just support IEEE 754 with no caveats? We wouldn't be the ones reaching. IEEE 754 is the standard that is reaching but its pretty well established (what, like 1985, right?).

(switching back to me) What known microprocessors and/or softfp implementations don't support NaN but do support IEEE 754?

specification/dsdl/architecture.tex
pavel-kirienko: Is explicit not better than implicit?

We state that one can't use anything except `[a-zA-Z0-9_]` for a namespace name. If the user wants to name their namespace after some obscure Internet startup, they will have to replace their dotcoms with underscore-coms manually.

thirtytwobits: okaaaaaaay

specification/dsdl/conventions.tex
pavel-kirienko: > shouldn't we just support IEEE 754 with no caveats?

But that's exactly what we do. I don't understand.

> What known microprocessors and/or softfp implementations don't support NaN but do support IEEE 754?

I don't know. Should we care? We refer to IEEE754 mostly for the sake of its well-defined bit-level representations. Platforms may internally use whatever floating point formats they want, it's none of our business, provided that they transform their internal formats into/from IEEE754 when serializing/deserializing.

specification/dsdl/conventions.tex
pavel-kirienko: There is a normative section that defines fields of type `bool`. Then there is a non-normative section (the one we're commenting under) that says that if you want a set of bit flags, you are advised to represent them in DSDL as a set of fields of type `bool` rather than a single field of type `uint*`.

thirtytwobits: okay.

specification/dsdl/conventions.tex
thirtytwobits: > But that's exactly what we do. I don't understand.

The caveat is "this pattern is discouraged because some nodes may not support IEEE 754 NaN". As far as I know NaN support isn't optional for IEEE754, correct?

pavel-kirienko: But support for IEEE 754 *is* very optional; hence support for NaN is also optional.

thirtytwobits: > IEEE 754 is very optional

Oh, sorry. Of course. So backing up then: mixed NaN support is a s**t-show. We should define and require support for a NaN value so nodes that do support IEEE754 know how to transmit them on the bus. Nodes that do not support IEEE754 will have to have a special case to handle them which is better then undefined behaviour. Perhaps we should add `sNaN` and `qNaN` as special, rational values in the language?

pavel-kirienko: How about this: I will add a few sentences requiring that platforms that do not natively support IEEE 754 must be able to accommodate IEEE 754 NaN values. This should probably go into "3.4.2 Primitive types" with an additional reference in "3.7.3.4 Floating point types". I don't think we should add NaN support for constant expressions yet, but this may be changed later without compromising backward compatibility.

thirtytwobits: but if I have a NaN value and I want to send it on the bus what do I do?

pavel-kirienko: You just send it. I am saying that we do not seem to need NaN values in DSDL.

```python
float16 a  # you can assign NaN to this at runtime
float16 A = NaN  # Can't do this
```

thirtytwobits: What about in dsdl expressions. What happens if we end up with something that is normally NaN like division by zero?

pavel-kirienko: We get a DSDL-processing-time error. The spec says that the division (and modulo) operator is not defined for x/0, so the DSDL front-end would just report an error.

specification/dsdl/conventions.tex
pavel-kirienko: I think I have addressed this in https://github.com/UAVCAN/specification/pull/46/commits/dbdc89b9d1b66f7ce292898bdb8e72df5a68a3aa

pavel-kirienko:
> The sentence is a little hard to parse. Is this version still correct:

Yes. Fixed in https://github.com/UAVCAN/specification/pull/46/commits/17ef29d15bbadb4e3a8e308a1eed07f96f949195

specification/dsdl/conventions.tex
pavel-kirienko: Fixed in https://github.com/UAVCAN/specification/pull/46/commits/f9ec8842207647366d35bc8d02a8f3c702f1da5d

pavel-kirienko:
I seem to have resolved all of the outstanding issues. Here is a compiled doc: [UAVCAN_Specification.pdf](https://github.com/UAVCAN/specification/files/3145880/UAVCAN_Specification.pdf)

(this PR is becoming very messy and hard to follow)

thirtytwobits:
👍

pavel-kirienko:
> :+1: 

@thirtytwobits LGTY?

thirtytwobits:
Yes. LGTM. (sorry)

PR #48, opened 2019-04-12
Proofreading of Chapter 3 - DSDL specification

pavel-kirienko:
This PR is against the changeset of #46. It contains minor corrections and fixes across the whole specification, mostly focused on minor clarifications and fixing incorrect term usage. The bulk of the changes is due to the automatic replacement of patterns like `node ID` to `node-ID`. I have also updated the DSDL rendering script to use PyDSDL v0.5.0 and fixed its version to prevent future breakage.

Originally I planned to postpone opening this PR until the previous one is accepted, but then I noticed that the old PR is likely to raise questions which are already addressed here, so it seemed sensible to come forward with my changes even if that might disrupt the workflow somewhat.

The changes should be relatively easy to review in the source form right here on GitHub. This is not a WIP; if we agree on the changes, they should be merged immediately.

### DIFF
--- a/specification/dsdl/architecture.tex
+++ b/specification/dsdl/architecture.tex
@@ -27,7 +27,8 @@ before the systems utilizing them are fielded.
 DSDL definitions can be used to automatically generate serialization (and deserialization) source code
 for any data type in a target programming language.
 A tool that is capable of generating serialization code based on a DSDL definition is called a \emph{DSDL compiler}.
-More generically, a software tool designed for working with DSDL definitions is called a \emph{DSDL processor}.
+More generically, a software tool designed for working with DSDL definitions is called a
+\emph{DSDL processing tool}.
 
 \subsection{Data types and namespaces}
 
@@ -113,12 +114,12 @@ provided that each version is defined at most once.
 Version number sequences can be non-contiguous,
 meaning that it is allowed to skip version numbers or remove existing definitions that are neither oldest nor newest.
 
-A data type definition may have an optional fixed port ID\footnote{Chapter \ref{sec:basic_concepts}.} value specified.
+A data type definition may have an optional fixed port-ID\footnote{Chapter \ref{sec:basic_concepts}.} value specified.
 
 The name of a data type definition file is constructed from the following entities
 joined via the ASCII dot character ``\verb|.|'' (ASCII code 46), in the specified order:
 \begin{itemize}
-    \item Fixed port ID in decimal notation, unless a fixed port ID is not provided for this definition.
+    \item Fixed port-ID in decimal notation, unless a fixed port-ID is not provided for this definition.
     \item Short name of the data type (mandatory, always non-empty).
     \item Major version number in decimal notation (mandatory).
     \item Minor version number in decimal notation (mandatory).
@@ -128,7 +129,7 @@ joined via the ASCII dot character ``\verb|.|'' (ASCII code 46), in the specifie
 \begin{figure}[H]
     $$
     \overbrace{%
-        \underbrace{\texttt{\huge{432}}}_{\substack{\text{fixed} \\ \text{port ID}}}%
+        \underbrace{\texttt{\huge{432}}}_{\substack{\text{fixed} \\ \text{port-ID}}}%
         \texttt{\huge{.}}%
     }^{\text{optional}}%
     \overbrace{%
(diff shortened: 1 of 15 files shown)


thirtytwobits:
I have notes in the PDF up through Section 3.4 so far. I'm trying  to finish  this today (my time). I guess I can't attach the PDF to this PR  so I'll go  back through and c&p the comments into the source when I'm done.

thirtytwobits:
Okay,  wow. Totally underestimated the size and information density of chapter 3.  First let me say, @pavel-kirienko ,  this is an impressive amount of work and, so far, seems to be well thought out. 

My new estimate for completing my review is not end-of-week (April  20th).

specification/dsdl/architecture.tex
```diff
@@ -27,7 +27,8 @@ \subsection{General principles}
 DSDL definitions can be used to automatically generate serialization (and deserialization) source code
 for any data type in a target programming language.
 A tool that is capable of generating serialization code based on a DSDL definition is called a \emph{DSDL compiler}.
-More generically, a software tool designed for working with DSDL definitions is called a \emph{DSDL processor}.
+More generically, a software tool designed for working with DSDL definitions is called a
+\emph{DSDL processing tool}.
```
thirtytwobits: Thanks for these definitions. The terminology is very helpful.

specification/dsdl/architecture.tex
```diff
@@ -189,25 +198,23 @@ \subsection{Elements of data type definition}
...
+unordered, possibly empty collection of \emph{constant attributes}.
 A data schema may describe either a \emph{structure object} or a \emph{tagged union object}.
-The value of a structure object is a function of the values of all of its fields.
-A tagged union object is formed from at least two fields,
-but it is capable of holding exactly one field at any given time.
-The value of a tagged union object is a function of which field it is holding at the moment and the value of said field.
+The value of a structure object is a function of the values of all of its field attributes.
+A tagged union object is formed from at least two field attributes,
+but it is capable of holding exactly one field attribute value at any given time.
+The value of a tagged union object is a function of which field attribute value
```
thirtytwobits: (suggested replacement for this sentence)
A tagged union's value is evaluated as a function of the tagged type the value is accessed through. There is no _intrinsic_ memo of an "actual" type associated with a union. Any such concept may, optionally, be provided by a type author using additional fields in a structure object that contains the tagged union.(footnote)

footnote - In this manner the DSDL tagged union mirrors the C language's concept of unions.

pavel-kirienko: > A tagged union's value is evaluated as a function of the tagged type the value is accessed through.

This would imply that the type of each option in a tagged union must be unique within that union. The following is valid according to the current draft, but not valid according to your proposed wording:

```python
@union
uint8 a
uint8 b
```

> There is no intrinsic memo of an "actual" type associated with a union.

The meaning of this sentence evades my understanding. Is it saying that unions are not types themselves?

pavel-kirienko: I am going to merge this now without your suggested change because I want to avoid merge conflicts in the target branch I'm currently editing. If you believe that this issue should be discussed further, please reply here or open a new ticket.

thirtytwobits: I'm being confused about the definition of `value` in this paragraph. In your mind, is the union's value the state of all the union's bits in memory (including undefined bits where a union was assigned using a field whose size is < the total size of the union) or is the value the representation obtained by accessing a union via a given field (i.e. ignoring undefined bits)?

pavel-kirienko: This is a more abstract statement than either of your options. "Value" of the union here is an abstract entity that affects the behavior or state of the process or logic it is used in. The implication of my definition, when considered in this setting, is that other fields of the union (that is, all fields except the active one) are irrelevant and do not affect the "behavior or state of the process or logic" which uses it. I avoided using any more specific definition because this is supposed to be a very abstract section. Should it be reworked?

thirtytwobits: meh. I think I might be bike-shedding now but I just don't like the use of the term "holding" (e.g. "...is a function of which field attribute value it is holding...") since that suggests some hidden mechanics that don't exist (i.e. "holding" is an active verb but unions are passive data constructs.

So, I'm okay with this I guess.

pavel-kirienko: We could replace "a function of which field attribute value it is holding" with "a function of which field attribute value is currently selected"? Or something like this. There is going to be another pass over the entire document when it's ready to weed out bad English and poorly phrased sentences (I'm not sure if Hamish is still available for this work; if not, we'll hire someone else).

thirtytwobits: Agreed. We'll leave this stuff for the tech writer.

hamishwillee:
> (I'm not sure if Hamish is still available for this work; if not, we'll hire someone else

@pavel-kirienko 

If you find a tech editor who has experience in standards documentation go for it. While I can certainly identify wording that isn't clear or is awkward, standards review requires a degree of precision that I don't normally have to aspire to. 

That said, I am available and willing to help over next month or so. Are you at the point where you need review? After that my contract changes and we would need to revisit.

pavel-kirienko:
@hamishwillee Thanks! We are *so very close* to the final review. Once this PR is merged (I think we're almost there), I am going to minorly restructure the transport chapter, add a few elaborations into the non-normative introductory chapter, and that would be it. I will let you know once we're ready.

PR #55, opened 2019-08-26
Various copy edits

thirtytwobits:


specification/application_layer/conventions.tex
```diff
@@ -3,7 +3,7 @@ \section{Application-level conventions}\label{sec:application_level_conventions}
 This section describes a set of high-level conventions designed to enhance compatibility
 of applications leveraging UAVCAN.
 The conventions described here are not mandatory to follow;
-however, every deviation should be justified and adequately documented.
+however, every deviation should be justified and adequately documented. % This may be outside the scope of this document, but what constitutes "adequately documented"? This may well be explained elsewhere.
```
pavel-kirienko:
```suggestion
however, every deviation should be justified and documented.
```

specification/application_layer/conventions.tex
```diff
@@ -86,21 +86,21 @@ \subsection{Rotation representation}
...
-        convention beforehand; such convention would be difficult to establish considering different
+        convention beforehand; a convention would be difficult to establish considering different
         demands of various use cases.
 
         \item Euler angles and fixed axis rotations typically cannot be used for computations directly
-        due to angular interpolation issues and singularities; thus, in order to make use of such
+        due to angular interpolation issues and singularities; thus, to make use of such
         representations, one often has to convert them to a different form (e.g., quaternion);
-        such conversions are relatively computationally heavy.
+        such conversions are relatively computationally heavy. % relative to what? Not sure we need word relatively here. May want to just say "such conversions can be computationally heavy"
```
pavel-kirienko: Indeed.
```suggestion
        such conversions are computationally heavy.
```

specification/application_layer/functions.tex
```diff
@@ -86,9 +86,10 @@ \subsection{Network-wide time synchronization}
 The detailed specification of the time synchronization algorithm is provided in the documentation
 for the message type \DSDLReference{uavcan.time.Synchronization}.
 
-An important service type \DSDLReference{uavcan.time.GetSynchronizationMasterInfo}
+An important service type \DSDLReference{uavcan.time.GetSynchronizationMasterInfo}
```
pavel-kirienko: (misconfigured text editor)
```suggestion
An important service type \DSDLReference{uavcan.time.GetSynchronizationMasterInfo}
```

specification/application_layer/functions.tex
```diff
@@ -86,9 +86,10 @@ \subsection{Network-wide time synchronization}
 The detailed specification of the time synchronization algorithm is provided in the documentation
 for the message type \DSDLReference{uavcan.time.Synchronization}.
 
-An important service type \DSDLReference{uavcan.time.GetSynchronizationMasterInfo}
+An important service type \DSDLReference{uavcan.time.GetSynchronizationMasterInfo} 
 is designed to provide nodes with information about the currently used time system
 and related data such as the number of leap seconds.
+% Would be better restated in active voice "\DSDLReference{uavcan.time.GetSynchronizationMasterInfo} is an important service type designed to provide nodes with information about the currently used time system and related data such as the number of leap seconds."
```
pavel-kirienko: Sure. Can we please incorporate this change here? Also, I see that the original phrase "an important service type" has no semantic weight and should be removed.

specification/application_layer/functions.tex
```diff
@@ -107,25 +108,26 @@ \subsection{Network-wide time synchronization}
 \subsection{Primitive types and physical quantities}
 
 The namespaces \DSDLReference{uavcan.si} and \DSDLReference{uavcan.primitive}
-included in the standard data type set are designed to provide a very generic and flexible,
-albeit neither bandwidth-efficient nor low-latency, method of real-time data exchange.
+included in the standard data type set are designed to provide a generic and flexible
+method of real-time data exchange; however, these namespaces are not bandwidth-efficient
```
pavel-kirienko: The properties of bandwidth utilization efficiency and latency are not defined on namespaces or data types contained therein. The new phrasing does not seem correct.

specification/application_layer/functions.tex
```diff
@@ -107,25 +108,26 @@ \subsection{Network-wide time synchronization}
...
+or low-latency.
 
 Generally, applications where the bus bandwidth and latency are important should minimize their reliance
 on these generic data types and favor more specialized types instead that are custom-designed for their
 particular use cases; e.g., vendor-specific types or application-specific types, either
 designed in-house, published by third parties\footnote{As long as the license permits.}, or supplied by
 vendors of COTS equipment used in the application.
 
-Vendors of COTS equipment should always ensure that at least some minimal functionality is available
+Vendors of COTS equipment should always ensure that some minimal functionality is available
```
pavel-kirienko: This is one of those issues in the original document I've meant to fix in the final pass over the text later. The level of requirement is higher than it should be, can we please fix it here while we're at it?

```suggestion
Vendors of COTS equipment are recommended to ensure that some minimal functionality is available
```

specification/application_layer/functions.tex
```diff
@@ -159,12 +161,12 @@ \subsubsection{Primitive namespace}
...
 The primitive types provide a very high degree of flexibility, but due to their lack of semantic information,
-their usage carries the risk of creating suboptimal interfaces that are difficult to use, validate, and scale.
+their use carries the risk of creating suboptimal interfaces that are difficult to use, validate, and scale.
 
-Normally, the usage of primitive types should be limited to very basic vendor-neutral interfaces for COTS
+Normally, the use of primitive types should be limited to very basic vendor-neutral interfaces for COTS
 equipment and software, debug and diagnostic purposes, and whenever there is a need to exchange data the
-type of which cannot be determined statically (an example of the latter use case is the register protocol
-described in section \ref{sec:application_register_interface}).
+type of which cannot be determined statically. An example of the latter use case is the register protocol
```
pavel-kirienko:
```suggestion
type of which cannot be determined statically.\footnote{An example of the latter use case is the register protocol
```

specification/application_layer/functions.tex
```diff
@@ -159,12 +161,12 @@ \subsubsection{Primitive namespace}
...
+described in section \ref{sec:application_register_interface}.
```
pavel-kirienko:
```suggestion
described in section \ref{sec:application_register_interface}.}
```

specification/application_layer/requirements.tex
```diff
@@ -7,10 +7,12 @@ \subsection{Port identifier distribution}
...
 
 The subject and service identifier values are segregated into three ranges:
-unregulated port identifiers that can be freely chosen by users and integrators (both fixed and non-fixed);
-regulated fixed identifiers for non-standard data type definitions
+\begin{itemize}
+    \item unregulated port identifiers that can be freely chosen by users and integrators (both fixed and non-fixed);
+    \item regulated fixed identifiers for non-standard data type definitions
 that are assigned by the UAVCAN maintainers for publicly released data types;
-and regulated identifiers of the standard data types that are an integral part of the UAVCAN specification.
+    \item and regulated identifiers of the standard data types that are an integral part of the UAVCAN specification.
```
pavel-kirienko:
```suggestion
    \item regulated identifiers of the standard data types that are an integral part of the UAVCAN specification.
```

specification/transport_layer/can.tex
```diff
@@ -2,14 +2,13 @@ \section{CAN bus transport layer specification}
 
 This section specifies the CAN-based transport layer of UAVCAN.
 
-Here and in the following parts of this section,
-"CAN" implies both CAN 2.0 and CAN FD, unless specifically noted otherwise.
+Throughout this section, "CAN" implies both CAN 2.0 and CAN FD, unless specifically noted otherwise.
```
pavel-kirienko:
```suggestion
Throughout this section, ``CAN'' implies both CAN 2.0 and CAN FD, unless specifically noted otherwise.
```

specification/transport_layer/can.tex
```diff
@@ -208,16 +207,15 @@ \subsubsection{Source node-ID field in anonymous message transfers}\label{sec:ca
...
+However, anonymous message transfers would violate this restriction because they don't have a unique node-ID.
 
 In order to work around this problem,
 UAVCAN requires that the source node-ID field of anonymous messages\footnote{Source node identifier
 is not defined for anonymous message transfers; see table \ref{table:common_transfer_properties}.}
 is initialized with a pseudorandom \emph{pseudo-ID value},
 and defines special logic for handling CAN bus errors during transmission of anonymous frames.
 
-The source of the pseudorandom data used for the pseudo-ID must be likely to produce different values
+The source of the pseudorandom data used for the pseudo-ID must be likely to produce different values % Not sure what 'must be likely' means in this context'
```
pavel-kirienko: Poor wording. What I am trying to say here is that the following conditions apply:

- Identical pseudo-ID values given different payloads are undesirable.
- Identical pseudo-ID values given identical payload are acceptable.

Perhaps the sentence should be changed to communicate these objectives?

specification/transport_layer/can.tex
```diff
@@ -368,9 +366,9 @@ \subsubsection{Multi-frame transfers}
 In the latter case, the last CAN frame will contain only the metadata,
 as specified below in this section.
 
-Each CAN frame of a multi-frame transfer except the last one
+Each CAN frame of a multi-frame transfer (except the last one)
```
pavel-kirienko: This is neither a minor addition nor an expansion but a crucial part of the description. Do we really want it parenthesized?

specification/application_layer/conventions.tex
pavel-kirienko: What we're doing here in this sentence is just advising the users on what the best practice should be. We want to ensure that their engineering approaches are solid and that if their implementations are re-used by other UAVCAN adopters, they do not stumble upon obscure compatibility issues. This is not really about the spec doc at hand; rather, it's about the larger ecosystem out there. Instead of trying to spec out everything we might want to release it in the current minimal form and then see if the ecosystem needs extra control input as we go.

specification/application_layer/functions.tex
pavel-kirienko: > system time

It's supposed to be ["time system"](https://en.wikipedia.org/wiki/Time#Time_conversions), not "system time". As in, an approach to timekeeping.

specification/transport_layer/can.tex
pavel-kirienko: Identical pseudo-ID given different payloads given simultaneous transmission triggers a collision error on the bus. Collision is bad because it forces all nodes involved in the collision to re-transmit, delaying their progress and temporarily disrupting all lower-priority traffic on the bus.

The anonymous transfer logic is fundamentally non-deterministic, but it doesn't mean that we don't care about its performance. Hence, we want to minimize the probability of collision, hence the objective to make the pseudo-ID sufficiently random.

specification/application_layer/conventions.tex
veronistar: Sounds like a good plan!

specification/application_layer/functions.tex
veronistar: Thanks for the clarification. Just flipping time and system around: 

Use uavcan.time.GetSynchronizationMasterInfo to provide nodes with information about the currently used time system and related data like the number of leap seconds added.

specification/transport_layer/can.tex
veronistar: Thanks for the explanation! I'm guessing that's already explained elsewhere in the document or known by whoever will implement this? 

If that's the case, I'm thinking this can just be rephrased to "The source of the pseudorandom data used for the pseudo-ID should ideally produce different values for different payloads. "

specification/application_layer/functions.tex
pavel-kirienko: Good. There's one last thing (I suppose I should mention this in the readme): some time ago Scott suggested that we avoid addressing the reader directly for reasons of consistency, so we probably shouldn't use the imperative mood like "use the data type".

specification/transport_layer/can.tex
pavel-kirienko: > I'm guessing that's already explained elsewhere in the document or known by whoever will implement this?

It is implied that the reader is already familiar with the CAN bus specification (it is an ISO standard), where this is explained.

> The source of the pseudorandom data used for the pseudo-ID should ideally produce different values for different payloads.

I am certain that everyone will understand what we mean, but I am not exactly sure about the technical correctness here. The number of distinct pseudo-ID values is less than the number of distinct payload values; therefore (pigeonhole principle) the ideal source of pseudo-ID, as we have defined it, cannot exist. The words "must be likely" in my original text were intended to address this problem, but I see now that they hardly added any clarity here. 

Maybe we could say something along these lines instead:

> The source of the pseudorandom data used for the pseudo-ID should be chosen so that the following principles are accommodated:
>
> - Identical pseudo-ID values given different payloads are undesirable.
> - Identical pseudo-ID values given identical payload are acceptable.

Or maybe I am just paying too much attention or otherwise overreacting. What do you think? What does @thirtytwobits think?

specification/application_layer/functions.tex
veronistar: how about just "uavcan.time.GetSynchronizationMasterInfo provides nodes with information about the currently used time system and related data like the number of leap seconds added." ?

specification/transport_layer/can.tex
thirtytwobits: I think what's confusing, this being part of a specification, is "must" coupled with "likely". The two concepts conflict (You MUST press the green button to avoid self-destruction since it is LIKELY that pressing the green button will cancel the self-destruct activity). We should change one of the two terms:

> If source of the pseudorandom data used for the pseudo-ID is likely to produce duplicate values more anonymous messages will be lost.

or 

> The source of the pseudorandom data used for the pseudo-ID must produce different values or more anonymous message will be lost.

pavel-kirienko: > The source of the pseudorandom data used for the pseudo-ID must produce different values **for different CAN frame payloads** or more anonymous messages will be lost.

The part about payloads seemed missing.

*(I don't see any contradiction in the green button example. Assuming that your self-destruction is certain unless the button is pressed; the chances of self-destruction, if the button is pressed, are less than certain; and pressing the button does not cause any unexpected side-effects, you MUST press the button (assuming that self-destruction is not your preferred outcome)).*

thirtytwobits:
Okay, so we seem to have consensus here. Who's actually changing what?

pavel-kirienko:
> Who's actually changing what?

Would you like to volunteer?

pavel-kirienko:
@thirtytwobits what I'm saying is that I could go ahead and update it myself, but it would be best if it was somebody who can write decent English (i.e., not me) to avoid new issues.

veronistar:
Is line 220 the last remaining open issue here? I recommend the following: 

"The source of the pseudorandom data used for the pseudo-ID must aim* to produce different values for different payloads."

* recommendation for a new footnote: "The number of distinct pseudo-ID values is less than the number of distinct payload values, but to avoid collision errors, this should be followed as much as possible."

pavel-kirienko:
> I recommend the following

This is sensible. I see no further issues with that excerpt.

> recommendation for a new footnote

It seems self-contradictory: "The number of distinct pseudo-ID values is less than the number of distinct payload values" means that collisions are unavoidable, but then "to avoid collision errors <...>" implies otherwise. We could bikeshed :bike: some more, or we could just drop the footnote and hope that the reader does not need to have this spelled out for them. The main text we were discussing earlier *is paramount to get right* because it actually defines behaviors; elaborations and clarifications are not critical and are cheap to drop. I think.

veronistar:
Let's go with removing the footnote - I'm all for removing unnecessary content.

pavel-kirienko:
@veronistar are you going to be updating this, or should we wait for @thirtytwobits?

veronistar:
@thirtytwobits confirmed he's good with merging this. I've just added a few commits for the outstanding comments. @pavel-kirienko - will you be doing the actual merging for this and PR #56?

pavel-kirienko:
@veronistar I see that the last three commits are made in a separate branch. Will you be cherry-picking them here?

> will you be doing the actual merging for this and PR #56?

Yes.

thirtytwobits:
Just waiting for Pavel to sign-off then I'll merge this. Closing #56 since I ended up pulling that change into this one.

pavel-kirienko:
@thirtytwobits could you please accept the seven pending suggestions I just un-resolved? I can't push to the origin.

specification/transport_layer/can.tex
thirtytwobits: ugh. we missed a comment

veronistar: :(
Not going to attempt to add a new commit, but found the new text in [this comment](https://github.com/UAVCAN/specification/pull/55#discussion_r330921220): 
"The source of the pseudorandom data used for the pseudo-ID must produce different values for different CAN frame payloads or more anonymous messages will be lost."

Issue #61, opened 2019-11-11
Data type extensibility


pavel-kirienko:
Thanks to the explicit syntax-semantics separation, type extensibility is less important for UAVCAN v1 than it is for some other protocols, such as v0 or MAVLink. However, it does not mean that it is not a major enabler for the long-term stability of the ecosystem.

We should explore the possible approaches to this problem and consider implementing one in a future minor revision of the protocol after v1.0, possibly v1.1.

I think we should be able to come up with a more rigorous and strongly-typed solution than naive field injection at the end. I have been thinking about the idea of polymorphism and publisher-subscriber type covariance for months, but the idea is still too vague for a constructive discussion, largely because I am still focused on more immediate issues. The discussion started by @tridge about the [adoption of v1 in ArduPilot](https://forum.uavcan.org/t/uavcan-v1-0-and-ardupilot/671) induced me to post a very brief and highly incomplete description of my ideas in that thread on the forum.

New proposals are welcome.

PR #64, opened 2019-11-15
v0 coexistence and data type extensibility

### LINKED ISSUE #59: Allow different data types under the same port-ID as long as they are semantically compatible and mutually bit-compatible
Consider replacing this in section 3.8.3.3:

> A port identifier of a given kind (subject or service) shall only be used with one major version of one data type in a given UAVCAN network.

Allow data types to be different as long as they are semantically compatible and mutually bit-compatible. See the rationale here https://forum.uavcan.org/t/uavcan-v1-0-and-ardupilot/671/8?u=pavel.kirienko:

image

### LINKED ISSUE #60: Formal mechanism for migrating a message definition between namespaces (renaming)
Context: https://forum.uavcan.org/t/uavcan-v1-0-and-ardupilot/671

> @thirtytwobits: Pavel, what do you think of defining a formal mechanism for migrating a message definition to a different namespace while maintaining compatibility with the existing one? Ultimately names shouldn’t matter on the wire so the only thing that prevents this is a lookup that can tolerate duplicate definitions when the duplicates are compatible.

> @pavel-kirienko: As I just posted above, nothing prevents one from migrating a data type from one namespace into another (unless you are using an experimental transport that uses the data type hash), but I don’t think the use cases where this would be actually needed are common.

> @tridge **[new inputs here, update your models]:** the reason for moving it is social/political. The original messages are in the namespaces of startups that don’t necessarily want their name to be ensconced forever in the message used across the industry. When they develop the message to solve a problem they are not thinking ahead to when they want it to get into an upstream implementation, they are just trying to hit a deadline for a shipping product. Later when they start caring about working with others we want to move responsibility for the message into a longer term organisation namespace.

This issue is the result of the above exchange. We need to mention in the specification that it's legal to rename a data type, including its relocation into a different namespace. Right now it's semi-illegal.

Note to our future selves: this change invalidates the proposal for the data type hash that I introduced here: https://forum.uavcan.org/t/alternative-transport-protocols/324. New proposals are welcome.

### LINKED ISSUE #63: v0/v1 coexistence
Consider modifying the CAN ID format to permit coexistence of v0/v1 on the same bus: https://forum.uavcan.org/t/uavcan-v1-0-and-ardupilot/671/22?u=pavel.kirienko

pavel-kirienko:
Fixes #59
Fixes #60
Fixes #63

Provides the necessary foundation for #61 to be added in the future (perhaps v1.1).

Serialization rules are modified as follows:

- Unused data left over after a serialized representation is deserialized shall be ignored.
- Missing data is substituted with zero bits.

The implicit truncation rule will allow us to introduce explicit support for subtyping in DSDL later without affecting backward compatibility. The implicit zero extension will enable us to introduce optional fields if that is found to be necessary, although as I described in #61 I hope to find a better solution than naive extensions. Together, the two rules shift UAVCAN on the strictness-extensibility spectrum far towards the latter end, thus resolving the extensibility and future-proofness issues discussed at https://forum.uavcan.org/t/uavcan-v1-0-and-ardupilot/671. Although this change comes at a cost for deeply isolated deployments which may not benefit from it, the cost is considered minor and the trade-off is in general beneficial for the ecosystem at large.

Data types are now explicitly allowed to be moved between namespaces and renamed. The validity of this used to be uncertain.

The transport layer implementation provides an explicit description of how to handle incoming transfers that are longer than the expected maximum. This is needed to support derived types and, shall we fail to find a better solution, extension fields. The transport spec is also explicit about short transfers not being an error.

The CAN ID layout has been modified to retain the old position of the node-ID field. This is suboptimal for the prospective migration to v2 in the far future, but it does solve the v0 coexistence problem and also provides the possibility to extend the service-ID field from the current 9 bits to 10 bits in the future if that is found to be necessary.

A pre-built PDF is attached: [UAVCAN_Specification.pdf](https://github.com/UAVCAN/specification/files/3851768/UAVCAN_Specification.pdf)


specification/dsdl/serialization.tex
```diff
@@ -77,12 +82,81 @@ \subsubsection{Implicit padding}
...
-as described in relevant sections of chapter \ref{sec:transport_layer};
-however, such padding behavior falls outside of the scope of the DSDL specification
-as it belongs to the domain of network transports rather than data presentation.
+\subsubsection{Implicit truncation of excessive data}
+
+When a serialized representation is deserialized, implementations shall silently\footnote{%
+    I.e., without registering an error of any kind.
+} ignore any excessive (unused) data remaining upon deserialization, irrespective of its amount.
+
+One of the consequences of the above requirement is that the transport layer will be able to introduce
```
thirtytwobits: This paragraph should use non-normative formatting.

thirtytwobits:
```suggestion
As a consequence of the above requirement the transport layer can introduce
```

specification/dsdl/serialization.tex
```diff
@@ -77,12 +82,81 @@ \subsubsection{Implicit padding}
...
+additional padding at the end of the serialized representation to satisfy its data size granularity constraints.
```
thirtytwobits:
```suggestion
additional padding at the end of a serialized representation to satisfy data size granularity constraints.
```

specification/dsdl/serialization.tex
```diff
@@ -77,12 +82,81 @@ \subsubsection{Implicit padding}
...
 
-Additionally, the transport layer may introduce extra padding at the end of the serialized representation,
-as described in relevant sections of chapter \ref{sec:transport_layer};
-however, such padding behavior falls outside of the scope of the DSDL specification
-as it belongs to the domain of network transports rather than data presentation.
+\subsubsection{Implicit truncation of excessive data}
+
+When a serialized representation is deserialized, implementations shall silently\footnote{%
+    I.e., without registering an error of any kind.
+} ignore any excessive (unused) data remaining upon deserialization, irrespective of its amount.
```
thirtytwobits:
```suggestion
} ignore any excessive (unused) data remaining upon deserialization.
```

specification/dsdl/serialization.tex
```diff
@@ -77,12 +82,81 @@ \subsubsection{Implicit padding}
...
+
+One of the consequences of the above requirement is that the transport layer will be able to introduce
+additional padding at the end of the serialized representation to satisfy its data size granularity constraints.
+The specifics of such padding behavior fall outside of the scope of the DSDL specification
+as they belong to the domain of network transports rather than data presentation.
+
+Another consequence is that a serialized representation constructed from an instance of type $B$ can be
```
thirtytwobits:
```suggestion
Because of Implicit truncation a serialized representation constructed from an instance of type $B$ can be
```

specification/dsdl/serialization.tex
```diff
@@ -77,12 +82,81 @@ \subsubsection{Implicit padding}
...
-When a serialized representation is deconstructed, the value of the trailing padding bits must be ignored.
+\subsubsection{Implicit zero extension of missing data}
+
+For the purposes of deserialization routines,
+the serialized representation of any instance of any data type shall be considered to contain an implicit
+sequence of trailing zero (0) bits of infinite length.
+
+Despite this rule, implementations are not allowed to intentionally truncate trailing zeros
+upon construction of a serialized representation of an object\footnote{%
+    Intentional truncation is prohibited because in a future revision of the specification the implicit zero
```
thirtytwobits:
```suggestion
    Intentional truncation is prohibited because in a future, major revision of the specification the implicit zero
```

specification/dsdl/serialization.tex
```diff
@@ -77,12 +82,81 @@ \subsubsection{Implicit padding}
...
+    extension rule may be removed.
+    If intentional truncation were allowed, removal of the implicit zero extension rule would have
+    broken backward compatibility.
+}.
+
+\begin{remark}
+    This example assumes that the reader is familiar with the variable-length array serialization rules,
+    explained later in this section.
```
thirtytwobits: can we link to that section here?

specification/dsdl/serialization.tex
```diff
@@ -77,12 +82,81 @@ \subsubsection{Implicit padding}
...
+    Let the data type $B$ be defined as follows:
+
+    \begin{minted}{python}
+        uint8[<256] array
+    \end{minted}
+
+    Then the serialized representation of $X$ can be deserialized into an instance of $B$ where the field
+    \verb|array| contains a sequence of four zeros: $0, 0, 0, 0$.
+
+    The implicit truncation rule is introduced to permit extension of data types by introducing additional fields
```
thirtytwobits: This paragraph seems misplaced. Should it appear at the start of this section?

specification/transport_layer/can.tex
```diff
@@ -33,26 +33,30 @@ \subsection{CAN ID structure}
...
             \\
 
             \textbf{Values} &
             \multicolumn{3}{c|}{$[0, 7]$} &
             $0$ &
             $\mathbb{B}$ &
-            \multicolumn{4}{c}{} &
+            $0$ &
+            \multicolumn{3}{c}{} &
             \multicolumn{12}{c|}{$[0, 32767]$} &
```
thirtytwobits: Why was this listed at a 15-bit value before? Was this an error?

specification/transport_layer/can.tex
```diff
@@ -33,26 +33,30 @@ \subsection{CAN ID structure}
             \multicolumn{4}{c|}{Service, not message} &
             \multicolumn{5}{c|}{Anonymous message} &
             \multicolumn{12}{c|}{\multirow{2}{*}{Subject-ID}} &
-            \multicolumn{8}{c|}{Protocol version}
-            \\\cline{2-4} \cline{7-10} \cline{23-29}
+            \multicolumn{1}{c|}{\multirow{2}{*}{R}} &
```
thirtytwobits: so how does this let us know the difference between a v0 and a v1 frame?

specification/transport_layer/transport_layer.tex
```diff
@@ -575,6 +575,72 @@ \subsection{Transfer-ID comparison}\label{sec:transfer_id_forward_distance}
...
+
+The deterministic nature of UAVCAN in general and DSDL in particular allows implementations to statically determine the
+maximum amount of memory that is required to contain a data object of a particular type.
+Consequently, an implementation that is interested in receiving data objects of a particular type\footnote{%
+    Messages, service requests, or service responses.
+}
+can statically determine the maximum length of the transfer payload.
+
+Implementations shall be able to handle incoming transfers containing a larger amount of payload data than expected.
+In the event of such extra payload being received, a compliant implementation should silently\footnote{%
```
thirtytwobits:
```suggestion
In the event of such extra payload being received, a compliant implementation should \footnote{%
```

specification/transport_layer/transport_layer.tex
```diff
@@ -575,6 +575,72 @@ \subsection{Transfer-ID comparison}\label{sec:transfer_id_forward_distance}
...
+\subsection{Payload truncation}
+
+The deterministic nature of UAVCAN in general and DSDL in particular allows implementations to statically determine the
+maximum amount of memory that is required to contain a data object of a particular type.
+Consequently, an implementation that is interested in receiving data objects of a particular type\footnote{%
+    Messages, service requests, or service responses.
+}
+can statically determine the maximum length of the transfer payload.
+
+Implementations shall be able to handle incoming transfers containing a larger amount of payload data than expected.
```
thirtytwobits:
```suggestion
Implementations shall handle incoming transfers containing a larger amount of payload data than expected.
```

specification/transport_layer/transport_layer.tex
```diff
@@ -575,6 +575,72 @@ \subsection{Transfer-ID comparison}\label{sec:transfer_id_forward_distance}
...
+In the event of such extra payload being received, a compliant implementation should silently\footnote{%
+    Such occurrence is not indicative of a problem so it shall not be reported as such.
+}
+discard the excessive (unexpected) data at the end of the received payload.
+The transfer CRC, if applicable, shall be validated regardless of the presence of the extra payload in the transfer.
+
+The requirement to silently discard the excessive payload data at the end of the transfer is motivated by
```
thirtytwobits:
```suggestion
The requirement to discard the excessive payload data at the end of the transfer is motivated by
```

specification/transport_layer/transport_layer.tex
thirtytwobits: how does a receiver determine the "true" payload size? Is this always determined via the MTU? Does this requirement limit portability to other network layers?

specification/transport_layer/transport_layer.tex
```diff
@@ -575,6 +575,72 @@ \subsection{Transfer-ID comparison}\label{sec:transfer_id_forward_distance}
...
+    \begin{minted}{python}
+        float32 parameter
+    \end{minted}
+
+    The payload truncation requirement guarantees that the two nodes will be able to interoperate despite
+    relying on incompatible data type definitions.
+    Under this example, the duty of ensuring the semantic compatibility lies on the system integrator.
+\end{remark}
+
+Implementations shall not enforce the minimum payload size on received transfers.
```
thirtytwobits: An example here for handling this case would be illuminating.

specification/dsdl/serialization.tex
pavel-kirienko: It doesn't have to be major because this change would not violate backward compatibility with existing systems.

specification/transport_layer/can.tex
pavel-kirienko: v0/v1 can be differentiated through the toggle bit, as I wrote here (search for the phrase "short answer"): https://forum.uavcan.org/t/uavcan-v1-0-and-ardupilot/671/9?u=pavel.kirienko.

specification/transport_layer/can.tex
pavel-kirienko: The field was 16-bit wide but the range was limited to [0, 32767], implicitly making the most significant bit reserved at zero. Now its reservation is expressed explicitly.

specification/transport_layer/transport_layer.tex
pavel-kirienko: The true payload size is determined by deserializing the DSDL object. We no longer have the tail array optimization in v1, so the (de)serialization process is completely decoupled from the transport layer. This does not affect portability in any way.

specification/dsdl/serialization.tex
pavel-kirienko: Sorry, this got messed up. Fixed.

specification/transport_layer/transport_layer.tex
pavel-kirienko: I just made a remark out of this with a reference to the DSDL serialization section.

pavel-kirienko:
> how do you know that a given frame is a v0 versus v1 frame

See the link I posted in the comments here. The short answer is that we look at the toggle bit.

> are we taking a hard dependency on a network having a statically defined MTU because of the payload truncation rule?

No. Could you please clarify which section may lead one to think that there is a dependency on the MTU?

-----

There is a subtle problem arising from the implicit zero extension rule: if a transport layer requires implicit padding, it shall be done with zero bits only. Otherwise, an implementation that attempts to deserialize an object using its subtype may consume the non-zero padding bits mistaking them for useful payload. This is somewhat suboptimal because of the stuff bits on CAN FD, but it is necessary for extensibility. I am going to update the PR with that change; we can revise it if necessary.

specification/dsdl/compatibility.tex
```diff
@@ -123,12 +123,63 @@ \subsubsection{Bit compatibility}
...
+    \end{minted}
+\end{remark}
+
 \subsubsection{Semantic compatibility}\label{sec:dsdl_semantic_compatibility}
 
 A data type $A$ is semantically compatible with a data type $B$
 if an application that correctly uses $A$ exhibits a functionally equivalent behavior to an application
 that correctly uses $B$.
-The property of semantic compatibility is commutative.
+The property of semantic compatibility is non-commutative.
```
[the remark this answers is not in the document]
pavel-kirienko: Suppose there is type A:

```py
void1
# Values above 100% are not allowed.
uint7 demand_factor_pct  # [percent]
```

Then there is type B:

```py
# Values above 100 indicate overload.
uint8 demand_factor_pct  # [percent]
```

A subscriber expecting B can accept A, but the reverse is not true.

specification/dsdl/compatibility.tex
```diff
@@ -180,7 +261,7 @@ \subsubsection{Versioning principles}
 
 In order to ensure a deterministic application behavior and ensure a robust migration path
 as data type definitions evolve, UAVCAN requires that all data type definitions that share the same
-full name and the same major version number must be semantically compatible with each other
+full name and the same major version number must be mutually semantically compatible with each other
```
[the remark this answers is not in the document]
pavel-kirienko: Defined as follows:

```tex
$A$ and $B$ are said to be \emph{mutually semantically compatible} if
$A$ is semantically compatible with $B$ and $B$ is semantically compatible with $A$.
```

specification/dsdl/compatibility.tex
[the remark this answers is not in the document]
thirtytwobits: gotcha. That might be a good example to add to the spec?

specification/dsdl/compatibility.tex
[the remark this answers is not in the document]
thirtytwobits: My concern here is that we only define this in a non-normative section (under section 3.8.3.4 using the PDF attached to this PR). Can we formally define this in a normative section before we use the concept in this normative section?

specification/dsdl/serialization.tex
thirtytwobits: Okay.

specification/transport_layer/can.tex
thirtytwobits: Oh. neat! (Not sure how I missed that). Okay, we need two add to non-normative notes to this PR so others don't miss this important part of the design:
1. In section 4.2.3.2 we need to call out that this is inverted from v0 and can be used now for v0<>v1 determination but in the future can be switched to allow v1<>v2 determination (i.e. only two major versions of UAVCAN can coexist on a single network).
2. Somewhere in this section, when talking about reserved bits, we should mention that the reserved bits cannot be used for protocol level detection and either reiterate the toggle bit requirements or link to that section.

specification/transport_layer/transport_layer.tex
thirtytwobits: If I'm deserializing a type then the type's total length is all I know at the DSDL processor level. I can accept additional frames in a multi-frame transfer using the tail-byte but to find additional bits within a frame I need some other length field. In CAN this is the MTU/dlc. Is there some other way you were thinking the "true" size of a transfer can be determined? Given this we are requiring an MTU to be available at the transport layer. What we cannot determine, however, is what is additional payload and what is empty overhead of the network frame (e.g. CAN-FD padding) but I think this is okay if we always include all padding in the CRC (as we do now). My comment here is, in your research into other technologies have you come across an interesting network that allows bits to be added to a frame (i.e. within the length specified by the MTU) after a sender has enqueued a message?

Additionally, there is a physical limit to to this rule that should be acknowledged as being unspecified. i.e. there is some maximum RX buffer length beyond which the system will reject the transfer. For CAN-FD this would be a system originally designed to accept only (say) 24 bytes attached to a network sending semantically compatible 64 byte messages (The CRC can be accumulated byte-wise so we aren't limited by the total number of frames in a transfer). The sender would have no way of knowing the receivers limitation. This means we require all nodes to support the same MTU in order to enable payload truncation. Note that this is a real possibility given memory-constrained MCUs where allocating 64-byte RX buffers is prohibitive and, for example, where the same limitation doesn't constrain the larger processor in the flight controller.

specification/dsdl/compatibility.tex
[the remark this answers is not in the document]
pavel-kirienko: Sorry, what I meant is that I just added that definition in this fresh commit here: https://github.com/UAVCAN/specification/pull/64/commits/8229d652c1aa344a545cf7e01a331edb92eae330. The definition is located in the normative body of the specification.

specification/transport_layer/can.tex
pavel-kirienko: Covered this in the new commit just now.

specification/dsdl/compatibility.tex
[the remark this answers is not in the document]
pavel-kirienko: Done.

specification/dsdl/serializable_types.tex
```diff
@@ -286,3 +286,41 @@ \subsubsection{Unions}\label{sec:dsdl_composite_tagged_unions}
...
+    \begin{minted}{python}
+        float64 a       # Index 0
+        int16[<=9] b    # Index 1
+        uint8 foo       # Index 2
+    \end{minted}
+
+    A message type that defines zero fields is a structural supertype of any other message type.
+    A message type may have an arbitrary number of supertypes as long as the field equality constraint is satisfied.
+
+    Polymorphic relations are not defined on service types and tagged unions.
```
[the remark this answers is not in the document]
pavel-kirienko: A type hierarchy is an ordered set of types such that for each pair of its members one type is a subtype of another,
and for any member its supertypes are located on the left.

Given a publisher P and a subscriber S, where P emits objects of type X, and S expects objects of type Y, we say that their publish-subscribe relation is:

- Type-covariant if X is a supertype of Y and X is not Y.
- Type-contravariant if X is a subtype of Y and X is not Y.
- Type-invariant if X is Y.
- Type-orthogonal if Y and X belong to different type hierarchies.

For example, a subscriber expecting `uavcan.primitive.Empty.1.0` is type-covariant with any publisher.

Due to the bidirectional nature of service exchanges, they can be either type-invariant or type-orthogonal, which renders structural subtyping much less useful. To work around that, we would have to reason about service requests and service responses as independent members of a type hierarchy. I am envisioning simple aliases for request and response schemas that we could add later, like `uavcan.file.Read.Request.1.0` (notice `Request`), each being an independent type. For now, I suggest to stash this into the TODO list and move on.

> Similarly, are you implying a prohibition on polymorphic types within tagged unions or just on the unions themselves?

I don't understand the difference between tagged unions and just unions. They are the same thing in my mind.

Regardless, this question forced me to think about this a bit more. The current prohibition derives from the common practice employed in major programming languages to not define polymorphic relations on unions (C++, D, Rust (enum or union)). UAVCAN, however, is different because it does not have any built-in means of dynamic polymorphism, and unions can be used to work around that (like variant types). It is useful to define subtyping such that a union is a subtype of another iff the type of each field of the subtype is of the same type or a subtype of the corresponding field in the supertype, the subtype contains all fields of the supertype at the same indexes, and the bit width of the union tag field is the same across the hiearchy. This is clunky but I will work out a cleaner definition for the spec. For example:

Base:

```python
@union
uavcan.primitive.Empty.1.0 foo
float16 bar
uint8 zoo
```

Subtype:

```python
@union
uavcan.diagnostic.Record.1.0 foo    # Subtype
float16 bar                         # Same
uint8 zoo                           # Same
int64[<=64] baz                     # New field
```

The general idea is to retain the full wire compatibility so that a subtype can be used to decode any instance of any supertype.

I am going to add this in my next commit and then we can review it.

---

I want to share a general observation I made while reviewing your feedback. The data type compatibility concepts as we have defined them no longer work with the new methods of ensuring compatible type extension -- namely, polymorphism, implicit truncation, and implicit zero extension. Same applies to versioning: addition of an optional field shall be allowed if its correct usage can be enabled by the implicit truncation/extension rules.

We shall discard the current data type compatibility policies and define a more compact set from scratch. I want to do that in a separate pull request, however.

specification/transport_layer/can.tex
```diff
@@ -317,16 +317,9 @@ \subsubsection{Tail byte}
 \subsubsection{Padding bytes}
 
 Certain transports (such as CAN FD) may not provide byte-level granularity of the CAN data field length.
-In that case, the useful payload is to be padded with the minimal number of padding bytes required
+In that case, the useful payload is to be padded with the minimal number of zero padding bytes required
 to bring the total length of the CAN data field to a value that can satisfy the length granularity constraints.
 
-When transmitting, each padding byte must be set to $\text{85} = \text{55}_\text{hex} = \text{0101\,0101}_\text{bin}$.
```
thirtytwobits: padding is padding. Why do we have to change it to be zero padding? Edges matter for wireless protocols too so we're reducing the performance for both CAN and for any packet-oriented wireless networks we want to support.

pavel-kirienko: Padding is not always padding. The implicit (zero) extension rule makes deserialization routines sensitive to the trailing unused data. For example, a publisher emits an object of type:

```py
uint16 foo
```

Imagine that the transport layer requires padding to 4 bytes, which is done with 0x55. Suppose that the published value is 0x1234, so the resulting serialized representation is `[0x34, 0x12, 0x55, 0x55]`.

Suppose that the receiving side wants to rely on the implicit zero extension rule and use the following definition:

```py
uint16 foo
uint16 bar
```

The expectation is that `foo` will be deserialized as 0x1234, and `bar` will be zero-extended as 0x0000. If arbitrary padding is allowed, the value of `bar` becomes undefined; in this particular example it would be 0x5555.

Therefore, any implicit extension rule requires that padding is done with the same-valued bytes as the implicit extension. In our case it's zero.

An alternative would be to replace the zero-extension rule with 0x55-extension rule, but its utility for data type designers is far lower.

thirtytwobits: Sorry, I should have been more clear: we can allow padding to be transmitted as non-zero as long as it is converted to zeros during deserialization...but then the data representation leaks into the driver...okay...and packet radios typically provide whitening capabilities (not unlike CAN's bit stuffing I suppose) so...never-mind. You're right.

specification/dsdl/serializable_types.tex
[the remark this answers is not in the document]
pavel-kirienko: See https://github.com/UAVCAN/specification/pull/64/commits/28902afd2aefa1e9e7d5b93aa1d642940b2bbdf7

specification/transport_layer/transport_layer.tex
pavel-kirienko: > Is there some other way you were thinking the "true" size of a transfer can be determined?

No. Here, the size only becomes known after deserialization, and it depends on the DSDL definition in use. As you wrote, the remaining unused data may be padding, sub-type fields, or both -- the receiver is unable to distinguish them.

> My comment here is, in your research into other technologies have you come across an interesting network that allows bits to be added to a frame (i.e. within the length specified by the MTU) after a sender has enqueued a message?

Not exactly, no. The closest behavior I am aware of is the min-size padding (like in Ethernet to guarantee a minimum packet propagation time for collision detection over long links) and cryptographic padding.

> Additionally, there is a physical limit to to this rule that should be acknowledged as being unspecified. i.e. there is some maximum RX buffer length beyond which the system will reject the transfer. For CAN-FD this would be a system originally designed to accept only (say) 24 bytes attached to a network sending semantically compatible 64 byte messages (The CRC can be accumulated byte-wise so we aren't limited by the total number of frames in a transfer). The sender would have no way of knowing the receivers limitation. This means we require all nodes to support the same MTU in order to enable payload truncation.

Yes, we will be forcing a trade-off upon system designers between extensibility and resource utilization. Computer networks typically expect the MTU to be the same across the link so the requirement is not new. Generally, you could perhaps see how an application that attempts to optimize RAM usage by limiting the MTU for a particular subject/service has a leaky abstraction spanning between the low-level network driver and the high-level data type definition.

pavel-kirienko: I just pushed a commit that I expect addresses this issue: https://github.com/UAVCAN/specification/pull/64/commits/b3893e367dbf9abe3058c1e07762f26dc20ee20d

> Additionally, there is a physical limit to to this rule that should be acknowledged as being unspecified

I did not acknowledge this explicitly. Instead, I worded this as a weak requirement ("should") with an elaboration about the motivation.

pavel-kirienko:
After this one is merged, the next step is to restructure the transport layer chapter (I started this on a separate branch already but I will need to start over because of the significant changes introduced here) and to simplify the data type compatibility rules.

specification/transport_layer/transport_layer.tex
```diff
@@ -575,6 +602,81 @@ \subsection{Transfer-ID comparison}\label{sec:transfer_id_forward_distance}
...
+The deterministic nature of UAVCAN in general and DSDL in particular allows implementations to statically determine the
+maximum amount of memory that is required to contain a data object of a particular type.
+Consequently, an implementation that is interested in receiving data objects of a particular type\footnote{%
+    Messages, service requests, or service responses.
+}
+can statically determine the maximum length of the transfer payload.
+
+Implementations shall handle incoming transfers containing a larger amount of payload data than expected.
+In the event of such extra payload being received, a compliant implementation should
+discard the excessive (unexpected) data at the end of the received payload\footnote{%
```
thirtytwobits: should we clarify that, for multi-frame transfers, the excessive data must be used to calculate the CRC before discarding it?

pavel-kirienko: Do I understand correctly that you want to replace or amend the following?

> The transfer CRC, if applicable, shall be validated regardless of the presence of the extra payload in the transfer.

pavel-kirienko:
I am going to merge this now to unblock my further progress. If you want to follow-up on https://github.com/UAVCAN/specification/pull/64#discussion_r361088885, let's register it as a separate issue.

PR #69, opened 2020-01-15
Timing recommendations without overspecification.

pavel-kirienko:
With this change, the only concrete time value remaining in the specification is the transfer-ID timeout (2 seconds).

Origin:
- https://github.com/UAVCAN/specification/pull/68#discussion_r365042541
- https://github.com/UAVCAN/specification/pull/68#discussion_r366159070

### DIFF
--- a/specification/transport/abstract.tex
+++ b/specification/transport/abstract.tex
@@ -626,7 +626,7 @@ The transmission timeout should be documented for each outgoing transfer port.
 
 \subsubsection{Pending service requests}
 
-In the case of cyclic transfer-ID transports (section \ref{sec:transport_transfer_id}),
+In the case of cyclic transfer-ID transports (section~\ref{sec:transport_transfer_id}),
 implementations should ensure that upon a transfer-ID overflow a service client session
 does not reuse the same transfer-ID value for more than one pending request simultaneously.
 
--- a/specification/transport/abstract.tex
+++ b/specification/transport/abstract.tex
@@ -610,6 +610,26 @@ A heterogeneous redundant transport shall meet \emph{either} of the following re
 
 \subsection{Transfer transmission}
 
+\subsubsection{Transmission timeout}
+
+The transport frames of a time-sensitive transfer whose payload has lost relevance due to
+its transmission being delayed should be removed from the transmission queue\footnote{%
+    Trailing transport frames of partially transmitted multi-frame transfers should be removed as well.
+    The objective of this recommendation is to ensure that obsolete data is not transmitted
+    as it may have adverse effects on the system.
+}.
+The time interval between the point where the transfer is constructed and the point where it is considered
+to have lost relevance is referred to as \emph{transmission timeout}.
+
+% "Output port" sounds better but this term is not defined.
+The transmission timeout should be documented for each outgoing transfer port.
+
+\subsubsection{Pending service requests}
+
+In the case of cyclic transfer-ID transports (section \ref{sec:transport_transfer_id}),
+implementations should ensure that upon a transfer-ID overflow a service client session
+does not reuse the same transfer-ID value for more than one pending request simultaneously.
+
 \subsubsection{Deterministic data loss mitigation}\label{sec:transport_deterministic_data_loss_mitigation}
 
 Performance of transport networks where the probability of a successful transfer delivery
--- a/specification/application/conventions.tex
+++ b/specification/application/conventions.tex
@@ -15,25 +15,15 @@ which is guaranteed to be above 127 for any transport.
 The two uppermost node-ID values are reserved for diagnostic and debugging tools;
 these node-ID values should not be used in fielded systems.
 
-\subsection{Timing constraints}
-
-\subsubsection{Recommended message publication timing constraints}
-
-Generally, it is recommended to abort a message transmission if it cannot be completed in 1 second.
-
-It is expected that high-frequency high-priority messages may opt for lower timeout values,
-whereas low-priority delayable data may opt for higher timeout values to account for network congestion.
-Refer to section~\ref{sec:transport_transfer_priority} on transfer prioritization.
-
-\subsubsection{Recommended service invocation timing constraints}
-
-It is recommended to abort a service transfer transmission if it cannot be completed in 1 second.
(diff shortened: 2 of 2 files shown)


specification/transport/abstract.tex
```diff
@@ -610,6 +610,26 @@ \subsubsection{Heterogeneous redundant transports}
...
+to have lost relevance is referred to as \emph{transmission timeout}.
+
+% "Output port" sounds better but this term is not defined.
+The transmission timeout should be documented for each outgoing transfer port.
+
+\subsubsection{Pending service requests}
+
+In the case of cyclic transfer-ID transports (section~\ref{sec:transport_transfer_id}),
+implementations should ensure that upon a transfer-ID overflow a service client session
+does not reuse the same transfer-ID value for more than one pending request simultaneously.
```
thirtytwobits: Just to be sure I understand; depending on how nodes implement their single-frame transfer logic, it could be that a node initiating 32 transfers in rapid succession and then failing to complete any transfers could cause the bus to become unresponsive until the transfer timeout expired for at least one of the pending transfers*. Specifically, we must require nodes receiving single frame transfers to ignore the state of their transfer queues** to prevent the bad, multi-frame sender from blocking all single-frame transfers as well. Even with this mitigation I'm just now realizing this is a failure-mode unique to multi-frame transfers which is not present for single-frame transfers. When designing a system this seems like a fairly significant point which requires critical messages to always be single-frame to retain the priority and reliability guarantees inherent in CAN in the face of misbehaving nodes.

\* I'm assuming a worst case here where all nodes on the bus are subscribed to the port ID for all 32 of the transfers.

\*\* This might seem obvious but an implementer seeking to avoid additional branches in their code may chose to treat single-frame transfers as a multi-frame transfer with a frame length of 1.

pavel-kirienko: I agree with the seemingly implied point that cyclic-TID tends to cause much headache and it's best to avoid it whenever possible -- the overflow state introduces many non-obvious edge cases that require careful handling. I don't think the following part is entirely correct though:

> a node initiating 32 transfers in rapid succession and then failing to complete any transfers could cause the bus to become unresponsive until the transfer timeout expired for at least one of the pending transfers

I think what you might be missing here is that transfer-ID spaces from different session specifiers are isolated from each other. Remember this new diagram?

image

What I tried to convey using the diagram (and the subsequent elaborations below it) is that a transfer is uniquely identified by the source and destination nodes (route specifier), its port identifier (data specifier), and the transfer-ID; you can model it as a three-dimensional space where a point defined by the triplet (data specifier, route specifier, transfer-ID) represents a particular unique transfer (or, in the case of cyclic-ID, it's only unique within the TID overflow period). So the scenario you described would not cause the bus to become unresponsive because the transfers emitted by other nodes over the same port would have a different route specifier.

I don't think the reasoning that follows holds because the base assumption seems incorrect.

thirtytwobits: Ah, you're right that my scenario would not affect all nodes, however; If the bad actor started 32 transfers on the same port to the same node then it would cause that one node to be unresponsive for the timeout period. This would extend a failure in the first node into the second iff the second treated single transfers as multi-frame transfers with a length of 1. This could be particularly concerning if the first node was a non-critical sub-system interfering with a critical one.
Because of this escalation the spec should specifically prohibit using the transfer ID for a single-frame transfer to keep these failure modes isolated.

pavel-kirienko: > then it would cause that one node to be unresponsive for the timeout period

I don't see why. The failure as described would affect not the entire receiving node but the specific *session* between the sending node and the receiving node. The failure state is confined to that one *session* and does not affect other activities on the network.

pavel-kirienko: This particular section we're discussing is about service requests. Suppose we have a client node that sends a service request transfer to a server node. The server application decided to take a nap but the client node is unaware of that. The client node sends a few more requests (all under the same service-ID) and while doing so it makes the TID overflow (say, it's sent 32 requests over CAN). The 32-rd request will use the same TID as the first one. Suppose that at this point, the server node wakes up and starts sending responses in the FIFO order; the first response will use the TID from the very first request, which (due to overflow) happens to be the same as in the very last request. Due to the overflow, the mapping from TID to local client states is no longer bijective, which renders the client unable to process the response.

If the recommendation given in this section is followed, the above failure mode will not occur. As an example, consider how this is handled in PyUAVCAN v1:

- [`pyuavcan.presentation.RequestTransferIDVariabilityExhaustedError`](https://pyuavcan.readthedocs.io/en/latest/.apidoc_generated/pyuavcan.presentation.html#pyuavcan.presentation.RequestTransferIDVariabilityExhaustedError)
- https://github.com/UAVCAN/pyuavcan/blob/58076a45277af1af4078ca3a90c05b672b4454ad/pyuavcan/presentation/_port/_client.py#L230-L231

In Libuavcan v0 this case is not handled at all, leaving it to the application to handle correctly (maybe it's bad): https://github.com/UAVCAN/libuavcan/blob/8913f14d11ceb558fdabc5cdbe1cc6a866ac4c67/libuavcan/src/node/uc_service_client.cpp#L62-L76

This is correct:

> The receiver is dumb so it will simply start new sessions relying on the transfer CRC as the final check?

But I think you really mean this:

> The receiver is dumb so it will simply start new **transfers** relying on the transfer CRC as the final check

pavel-kirienko: > If the receiver is "dumb" what constitutes an overflow for it?

This is something that was dictated very unambiguously before my Transport Layer rewrite; remember there was a page full of pseudocode? In the rewrite, I avoided dictating any specific logic because while working on PyUAVCAN I noticed that it might be too constraining, especially considering that we are planning to support other transports in the future. Instead, the new spec simply requires that each unique transfer is processed once; how exactly that is ensured is implementation-defined. I think this abstract approach is superior because the specification should not micro-manage irrelevant implementation details that do not affect externally observable behaviors.

A receiving node normally does not need to handle overflows. To comply with the requirements, it is sufficient to compare the TID of a newly received frame against the TID of a previously received frame.

thirtytwobits: okay. Thanks for staying with me here.

Issue #70, opened 2020-02-26
Switch the bit order


pavel-kirienko:
Context: https://forum.uavcan.org/t/big-endian-vs-little-endian-in-the-context-of-bit-level-encoding/428/14?u=pavel.kirienko

Affected sections:

3.7.1.2 Bit and byte ordering
3.7.3.1 General principles
3.7.5 Composite types
3.8.3.4 Data type version selection (the bit flags in the example should be ordered the other way)
4.2.3 Examples

PR #76, opened 2020-04-06
Start of #70

thirtytwobits:
This will be a sliding review. I wanted to get this to you @pavel-kirienko sooner so you could comment as I go along. So far I've made it to section 3.7.4 and will continue with section 3.7.5 next and so on.

### DIFF
--- a/specification/dsdl/serialization.tex
+++ b/specification/dsdl/serialization.tex
@@ -353,7 +353,7 @@ of $b$ ensures byte alignment of this implicit field when serialized\footnote{Fu
 may allow this second step to be modified but the default action will always be to byte-align the implicit
 length field.}:
 
-$$2^{\lceil{}\log_2 (max(8, b))\rceil{}}$$
+$$2^{\lceil{}\log_2 (\text{max}(8, b))\rceil{}}$$
 
 The number of elements $n$ contained in the variable-length array is encoded
 in the serialized representation of the implicit length field
--- a/specification/dsdl/serialization.tex
+++ b/specification/dsdl/serialization.tex
@@ -375,7 +375,7 @@ a fixed-length array of $n$ elements.
         @assert _offset_.max / 8 <= 7  # This would fail.
     \end{minted}
 
-    In the above example the author attempted to fit the message into a single CAN 2.0 frame but
+    In the above example the author attempted to fit the message into a single Classic CAN frame but
     did not account for the implicit length field. The correct version would be:
 
     \begin{minted}{python}
--- a/specification/dsdl/serialization.tex
+++ b/specification/dsdl/serialization.tex
@@ -390,9 +390,9 @@ a fixed-length array of $n$ elements.
     \begin{minted}{python}
         uint8 first
         uint8 implicit_length_field  # Set to 3, because the array contains three elements
-        AnyType.1.0 item_0
-        AnyType.1.0 item_1
-        AnyType.1.0 item_2
+        uint8 item_0
+        uint8 item_1
+        uint8 item_2
     \end{minted}
 \end{remark}
 
--- a/specification/dsdl/serialization.tex
+++ b/specification/dsdl/serialization.tex
@@ -366,7 +366,7 @@ a fixed-length array of $n$ elements.
 
 \begin{remark}
     Datatype authors must take into account that dynamic arrays with a capacity of $<=255$ elements will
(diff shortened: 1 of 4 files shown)


specification/dsdl/serialization.tex
```diff
@@ -335,21 +365,31 @@ \subsubsection{Variable-length array types}\label{sec:dsdl_serialized_variable_l
...
+    consume an additional 8-bits of the message (where a capacity of $<=65535$ will consume 16-bits and so on).
+    For example:
+
+    \begin{minted}{python}
+        uint8 first
+        AnyType.1.0[<6] second         # The implicit length field is 8 bits wide
+        @assert _offset_.max / 8 <= 7  # This would fail.
+    \end{minted}
+
+    In the above example the author attempted to fit the message into a single CAN 2.0 frame but
```
pavel-kirienko, line 378: The term Classic CAN is used in the CAN FD specification.

```suggestion
    In the above example the author attempted to fit the message into a single Classic CAN frame but
```

specification/dsdl/serialization.tex
```diff
@@ -335,21 +365,31 @@ \subsubsection{Variable-length array types}\label{sec:dsdl_serialized_variable_l
...
+    did not account for the implicit length field. The correct version would be:
 
     \begin{minted}{python}
-        void2                        # Padding - good practice but not required
-        AnyType.1.0[<42] array       # The implicit length field is 6 bits wide
-        @assert _offset_.min == 8    # Ensuring that the first element is byte-aligned
+        uint8 first
+        AnyType.1.0[<5] second         # The implicit length field is 8 bits wide
```
pavel-kirienko, line 383:
```suggestion
        uint8[<=5] second              # The implicit length field is 8 bits wide
```

specification/dsdl/serialization.tex
```diff
@@ -335,21 +365,31 @@ \subsubsection{Variable-length array types}\label{sec:dsdl_serialized_variable_l
...
+    be equivalent to that of the following definition:
 
     \begin{minted}{python}
-        void2                        # Padding - good practice but not required
-        uint6 implicit_length_field  # Set to 3, because the array contains three elements
+        uint8 first
+        uint8 implicit_length_field  # Set to 3, because the array contains three elements
         AnyType.1.0 item_0
         AnyType.1.0 item_1
         AnyType.1.0 item_2
```
pavel-kirienko, line 395:
```suggestion
        uint8 item_0
        uint8 item_1
        uint8 item_2
```

specification/dsdl/serialization.tex
```diff
@@ -119,16 +119,16 @@ \subsubsection{Implicit truncation of excessive data}
 
 \subsubsection{Implicit zero extension of missing data}
 
-For the purposes of deserialization routines,
-the serialized representation of any instance of any data type shall be considered to contain an implicit
-sequence of trailing zero (0) bits of infinite length.
+For the purposes of deserialization routines, if any received message is shorter than the known datatype an
+implementation shall attempt to accept this message by filling the missing data with zeros (0)\footnote{%
+    When accepting multi-part transfers the subsequent evaluation of the checksum may fail in such cases.
```
thirtytwobits, line 124: How about this?

```suggestion
For the purposes of deserialization routines, if any received message is shorter than the known datatype an
implementation shall attempt to accept this message in a way that does not preclude integrity checking on
received multi-frame transfers\footnote{%
    This can be implemented by checking for out-of-bounds access during deserialization and returning zeros
    if an out-of-bounds access is detected. This is where the name "implicit zero extension rule" is derived
    from.
}.
```

pavel-kirienko: Maybe it's best to go back to the original phrasing? I think this new wording you proposed is too involved with the transport layer, which is bad because in this chapter we're talking about DSDL. We could keep your footnote though.

If you want to keep the footnote, please fix the quotes:

    ``implicit zero extension rule''

or `\emph{implicit zero extension rule}`.

a1ien:
Forgot fix section 3.7.5?
image

a1ien:
Also bit sequence in comment for first value incorrect. uint12 but len 16

pavel-kirienko:
Hi Ilya,

This PR brings in the changes up to section 3.7.4 only, as Scott said in
the OP comment. You are commenting on section 3.7.5 which is intentionally
not updated yet.

The comment you are referring to is correct because it represents the value
before truncation.

Issue #49, opened 2019-04-16
Add section numbers to pdf bookmarks


thirtytwobits:
It'd be helpful to have the section numbers in the index. As the below image shows, it's not easy to do something like "go to section 3" without them:

![Screen Shot 2019-04-16 at 10 21 48 AM](https://user-images.githubusercontent.com/19180861/56230595-7ab72e00-6031-11e9-9401-008218293399.png)

PR #84, opened 2020-06-08
Make Section Numbers Appear in PDF Bookmark Tree

Finwood:
By setting hyperref's `bookmarksnumbered` option, the generated pdf bookmark tree has section numbers prepended to the entries.

image

This would close UAVCAN/specification#49.

### DIFF
--- a/specification/uavcandoc.cls
+++ b/specification/uavcandoc.cls
@@ -275,6 +275,9 @@
 \preto\listoftables{\hypersetup{linkcolor=black}         \setlength{\parskip}{0em}}
 \appto\listoftables{\hypersetup{linkcolor=colorhyperlink}\setlength{\parskip}{\DefaultParskip}}
 
+% Include section numbers in the pdf bookmark tree.
+\hypersetup{bookmarksnumbered}
+
 %
 % Table macros.
 %


clyde-johnston:
I am not clear where this appears in the document. I think it appears in the sidebar of the PDF viewer and allows the user to quickly jump to the section or page he/she wants. This is a useful feature but its presentation is PDF viewer specific; the publisher has no control over this.

Normally with a list like this, though, the space between the number (or label) would be variable and act as padding so that the list item (i.e. the description) would all align vertically. You notice that, for example, "Introduction" and "Basic concepts" in items 1 and 2 are not aligned. If there were a 3.10, this misalignment would be even more apparent.

In my PDF viewer, hyperlinked ToC lists appear like this:

image

In a well typeset document, the table of contents would look like this:
ToC

As for the content, I do not see any problems.

Finwood:
> I am not clear where this appears in the document. I think it appears in the sidebar of the PDF viewer and allows the user to quickly jump to the section or page he/she wants.

That is correct, this PR adds the previously missing section numbers to the existing sidebar table-of-contents tree.

> its presentation is PDF viewer specific; the publisher has no control over this

This is not entirely true. The graphical rendering of the tree is surely platform and viewer specific, the text however is not. Viewers do not automagically add section numbers to the bookmark text, allowing publishers greater control over the ToC tree's appearance.
This also applies to your example, apparent in the first entry _Table 1. Device Summary_ not having a number prepended.

> In a well typeset document, the table of contents would look like this:

Don't we have that already? :thinking: 

image

clyde-johnston:
>This is not entirely true. The graphical rendering of the tree is surely platform and viewer specific, the text however is not. 

Hmmm....  :confused: 

I did say "_its presentation_ is PDF viewer specific; the publisher has no control over this". The content/text is another matter. :eyeglasses: 

>Don't we have that already? thinking

I had not seen the document; just the snippet presented in this PR.

I would have more than a few quibbles with the ToC presentation though; particularly with the change in font sizes (the page numbers look terrible!). Maybe you should remove the third-level items from the list. The reader is unlikely to be searching for these details in the ToC anyway and it would improve readability. If you do need it, then I would make the font size the same as the second-level items.

Anyway, it is certainly not as clear and well presented as the original example. There it uses white space and emphasis well, making it easy to read. You might want to use it as a model for how to present table of contents well.

Finwood:
> I would have more than a few quibbles with the ToC presentation though; particularly with the change in font sizes (the page numbers look terrible!). Maybe you should remove the third-level items from the list.

I agree, that _does_ look rather strange in my opinion as well. This is, however, outside the scope of this PR.

The three lines I would like to be merged address solely the section numbers in pdf bookmarks, as requested by @thirtytwobits in #49. If my attempt at solving _that_ issue is not sufficient, please tell me why, and I'll be happy to try adapting my solution. If the changes in #49 are indeed not wanted by this project's maintainers, please just drop this PR ¯\_(ツ)_/¯

> You might want to use it as a model for how to present table of contents well.

Since this single commit adding three single lines is my first contribution to this document, I don't see myself being in a position deciding on layout and typesetting. Design is a matter of personal taste, after all.

clyde-johnston:
I have no problem with the hyperlink bookmarks.  I think this is a useful feature for the reader so I would go ahead and merge this PR.

On the question that I was asked:
> @clyde-johnston does it look ok to you?

The short answer is: no (if by "it" you mean the document).  I would say, after now successfully compiling the document, that it suffers the same problems as the Zubax datasheets.  In fact, it looks like uavcandoc.cls has much of the same code as zubaxdoc.cls.  We should probably address the problems in both (e.g. title formats on later versions) together.

PR #85, opened 2020-06-11
Backport fixes from zubaxdoc.cls to uavcandoc.cls

clyde-johnston:
This PR fixes issue #35.

### DIFF
--- a/specification/dsdl/attributes.tex
+++ b/specification/dsdl/attributes.tex
@@ -54,7 +54,7 @@ the value of its constant attribute.
 The set of compatible types depends on the type of the initialized constant attribute,
 as specified in table~\ref{table:dsdl_constant_init_pattern}.
 
-\begin{UAVCANSimpleTable}{Permitted constant attribute value initialization patterns}{|l | X | X[2] | X[2]|}
+\begin{UAVCANSimpleTable}[wide]{Permitted constant attribute value initialization patterns}{|l | X | X[2] | X[2]|}
     \diagbox[font=\footnotesize]{Constant\\type\\category}{Expression\\type} &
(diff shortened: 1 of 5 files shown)


pavel-kirienko:
The tables in the document have to be updated because now with the new default left skip behavior some of them should be amended with `[wide]` (those that are wide enough).

The doc should also be skimmed to look for formatting issues introduced by the switch to newer LaTeX.

I think we should also release a new version of the texer container but this is obviously out of the scope of this PR.

clyde-johnston:
> I think we should also release a new version of the texer container but this is obviously out of the scope of this PR.

Opened issue [#9](https://github.com/UAVCAN/docker_toolchains/issues/9) in UAVCAN/docker_toolchains for this.

clyde-johnston:
> The tables in the document have to be updated because now with the new default left skip behavior some of them should be amended with `[wide]` (those that are wide enough).
> 
> The doc should also be skimmed to look for formatting issues introduced by the switch to newer LaTeX.

This should be directed to the typographer. My quick scan of the document did not find any issues although tables 4.2 and 4.3 might be candidates for the [wide] option.

Also, some tables seem to have been included as images and thus to do not conform to the standard alignments. It might be better to change these to LaTeX tables.

pavel-kirienko:
Okay. I looked through the doc and I think we need `[wide]` on tables 3.12, 3.14, 4.2, 4.3, 7.1, 7.5, 7.6, plus all tables generated by `render_dsdl.py` (they are raw `tabu`s so maybe they don't need fixing at all). Could you please update this PR accordingly?

clyde-johnston:
> Could you please update this PR accordingly?

Done.

PR #87, opened 2020-06-24
Fixing bit order for example in section 3.8.3.4

thirtytwobits:
Small fix.


thirtytwobits:
> Conventions dictate that we should put comments under the definition, not above.

Which comments are you referring to? The ones I added were not documenting a data field. They were general comments within the type.
 
> Per #70, 3.7.5 and 4.2.3 are next?
yes.

pavel-kirienko:
Here's one example; it was like this:

```
bool flag_cooling_system_b_active
bool flag_cooling_system_a_active
# Status flags (this is the recommended practice).
```

Now it's like this:

```
# Status flags (this is the recommended practice).
bool flag_cooling_system_a_active
bool flag_cooling_system_b_active
```

Issue #81, opened 2020-05-23
UAVCAN/CAN PHY: Mention the recommended impedance


pavel-kirienko:
See for context: https://forum.uavcan.org/t/on-can-cable-impedance-and-rj45-connectors/845

thirtytwobits:
Super-ceded by removal of the physical layer from the specification.

PR #95, opened 2020-07-29
removing physical layer section from spec

thirtytwobits:
As discussed on the dev call, the physical layer specification for CAN
transports is significantly less mature than the rest of this specification and is
also the most difficult layer to standardize in generic terms. Proper
functioning of the CAN data-link layer requires properly designed
electrical busses with properties specific to the physical
dimensions of a vehicle, the topology of the network, the number of transceivers connected
to the network, the location of the transceivers in relation to each other,
and the electrical properties of each transceiver. As high data-rate
variants like CAN-FD 1/5Mbs are deployed this electrical engineering
will become even more important even for very small CAN busses.
Because of this we are removing the physical layer specification from
the core document but reserve the right to create sibling documents
that standardize physical UAVCAN networks based on vehicle morphology
and supported transports.


pavel-kirienko (approved):
Four weeks with no objections. I think this is safe to merge.

This PR invalidates https://github.com/UAVCAN/specification/issues/81

Issue #90, opened 2020-07-13
Delimited serialization


pavel-kirienko:
https://forum.uavcan.org/t/data-type-extensibility-and-composition/827/13?u=pavel.kirienko

PR #97, opened 2020-09-11
Beta PR 2 of 2: delimited serialization

pavel-kirienko:
Fixes https://github.com/UAVCAN/specification/issues/90

Context, motivation, discussion: https://forum.uavcan.org/t/data-type-extensibility-and-composition/827

Implementation: https://github.com/UAVCAN/pydsdl/pull/45; its documentation: https://pydsdl--45.org.readthedocs.build/en/45/

-----

This is the last PR separating us from v1.0-beta; you will notice that I have renamed the document into "**Specification v1.0-beta**" in this branch already.

The changeset is large because it required updating many independent sections of the document. I recommend reviewing it by simply reading chapter **3 DSDL**, without any regard for the diffs. Afterward, the second pass can be done by looking at the diffs only.

There are some inconsequential changes in the script `render_dsdl.py` that are safe to ignore. This script is seriously getting out of hand and I wish someone could rewrite it using a proper template engine like Jinja. I never expected it to grow so complex.

I also removed the term "data schema" from the spec because it did not reflect the reality accurately and was used incorrectly in several places. Instead, we now use "data type". This change made it painfully obvious that service types should not even belong in the category of serializable types because they are not serializable, but I suppose we'll not be changing this detail now unless someone is willing to volunteer.

### DIFF
--- a/specification/dsdl/architecture.tex
+++ b/specification/dsdl/architecture.tex
@@ -6,7 +6,7 @@ In accordance with the UAVCAN architecture, DSDL allows users to define data typ
 message types and service types.
 Message types are used to exchange data over publish-subscribe one-to-many message links identified by subject-ID,
 and service types are used to perform request-response one-to-one exchanges (like RPC) identified by service-ID.
-A service type is composed of exactly two unutterable types:
+A service type is composed of exactly two unutterable inferior data types:
 one of them is the request type (its instances are transferred from client to server),
 and the other is the response type (its instances are transferred from the server back to the client).
 
@@ -93,7 +93,7 @@ A name component shall not match any of the reserved word patterns,
 which are listed in table~\ref{table:dsdl_reserved_word_patterns}.
 
 The length of a full data type name shall not exceed 50
-characters\footnote{This includes the name component separators.}.
+characters\footnote{This includes the name component separators, but not the version.}.
 
 Every data type definition is assigned a major and minor version number pair.
 In order to uniquely identify a data type definition, its version numbers shall be specified.
@@ -228,11 +228,17 @@ such as deprecation markers (notifying its users that the data type is no longer
 or assertions (special statements introduced by data type designers
 which are statically validated by DSDL processing tools).
 
-A service type definition is a special case: it can be neither a structure nor a tagged union,
-and it does not contain attributes;
-instead, it directly describes two inferior data types:
-the request type and the response type, separated by the \emph{service response marker},
-which are otherwise ordinary data types.
+Service type definitions are a special case:
+they cannot be instantiated or serialized, they do not contain attributes,
+and they are composed of exactly two inferior data type definitions\footnote{
+    A service type can be thought of as a specialized namespace that contains two types and
+    has some of the properties of a type, such as name and version.
+}.
+These inferior types are the service request type and the service response type,
+separated by the \emph{service response marker}.
+The inferiors are otherwise ordinary data types except that they are unutterable\footnote{Cannot be referred to.}
+and they derive some of their properties\footnote{Like version numbers or deprecation status.}
+from their \emph{parent service type}.
 
 \subsection{Serialization}
 
@@ -256,4 +262,5 @@ which is attached to the serialized object, along with other metadata, in a mann
 transport layer specification (chapter~\ref{sec:transport}).
 For more information on port identifiers and data type mapping refer to section~\ref{sec:basic_subjects_and_services}.
 
-The bit length set is not defined on service types (only on their inferiors: request and response).
+The bit length set is not defined on service types (only on their inferiors: request and response)
+because they cannot be instantiated.
--- a/specification/dsdl/architecture.tex
+++ b/specification/dsdl/architecture.tex
@@ -6,7 +6,7 @@ In accordance with the UAVCAN architecture, DSDL allows users to define data typ
 message types and service types.
 Message types are used to exchange data over publish-subscribe one-to-many message links identified by subject-ID,
 and service types are used to perform request-response one-to-one exchanges (like RPC) identified by service-ID.
-A service type is composed of exactly two unutterable inferior data types:
+A service type is composed of exactly two inner data types:
 one of them is the request type (its instances are transferred from client to server),
 and the other is the response type (its instances are transferred from the server back to the client).
 
@@ -230,13 +230,13 @@ which are statically validated by DSDL processing tools).
 
 Service type definitions are a special case:
 they cannot be instantiated or serialized, they do not contain attributes,
-and they are composed of exactly two inferior data type definitions\footnote{
+and they are composed of exactly two inner data type definitions\footnote{
     A service type can be thought of as a specialized namespace that contains two types and
     has some of the properties of a type, such as name and version.
 }.
-These inferior types are the service request type and the service response type,
+These inner types are the service request type and the service response type,
 separated by the \emph{service response marker}.
-The inferiors are otherwise ordinary data types except that they are unutterable\footnote{Cannot be referred to.}
+They are otherwise ordinary data types except that they are unutterable\footnote{Cannot be referred to.}
 and they derive some of their properties\footnote{Like version numbers or deprecation status.}
 from their \emph{parent service type}.
 
@@ -262,5 +262,5 @@ which is attached to the serialized object, along with other metadata, in a mann
 transport layer specification (chapter~\ref{sec:transport}).
 For more information on port identifiers and data type mapping refer to section~\ref{sec:basic_subjects_and_services}.
 
-The bit length set is not defined on service types (only on their inferiors: request and response)
+The bit length set is not defined on service types (only on their request and response types)
 because they cannot be instantiated.
--- a/specification/dsdl/architecture.tex
+++ b/specification/dsdl/architecture.tex
@@ -6,10 +6,9 @@ In accordance with the UAVCAN architecture, DSDL allows users to define data typ
 message types and service types.
 Message types are used to exchange data over publish-subscribe one-to-many message links identified by subject-ID,
 and service types are used to perform request-response one-to-one exchanges (like RPC) identified by service-ID.
-A message data type defines one data schema of the message object,
-and a service data type contains two independent data schema definitions:
-one of them applies to the request object (transferred from client to server),
-and the other applies to the response object (transferred from the server back to the client).
+A service type is composed of exactly two unutterable types:
+one of them is the request type (its instances are transferred from client to server),
+and the other is the response type (its instances are transferred from the server back to the client).
 
 Following the deterministic nature of UAVCAN, the size of a serialized representation of any
 message or service object is bounded within statically known limits.
@@ -194,16 +193,15 @@ nesting\footnote{%
     \end{figure}
 \end{remark}
 
-\subsection{Elements of data type definition}
+\subsection{Elements of data type definition}\label{sec:dsdl_elements_of_data_type_definition}
 
 A data type definition file contains an exhaustive description of a particular version of the said data type in the
 \emph{data structure description language} (DSDL).
-As explained above, one data type definition contains either one or two data schema definitions,
-for message types and service types, respectively.
 
-A data schema definition contains an ordered, possibly empty collection of \emph{field attributes} and/or
+A data type definition contains an ordered, possibly empty collection of \emph{field attributes} and/or
 unordered, possibly empty collection of \emph{constant attributes}.
-A data schema may describe either a \emph{structure object} or a \emph{tagged union object}.
+
+A data type may describe either a \emph{structure object} or a \emph{tagged union object}.
 The value of a structure object is a function of the values of all of its field attributes.
 A tagged union object is formed from at least two field attributes,
 but it is capable of holding exactly one field attribute value at any given time.
@@ -230,6 +228,12 @@ such as deprecation markers (notifying its users that the data type is no longer
 or assertions (special statements introduced by data type designers
 which are statically validated by DSDL processing tools).
 
+A service type definition is a special case: it can be neither a structure nor a tagged union,
+and it does not contain attributes;
+instead, it directly describes two inferior data types:
+the request type and the response type, separated by the \emph{service response marker},
+which are otherwise ordinary data types.
+
 \subsection{Serialization}
 
 Every object that can be exchanged between UAVCAN nodes has a well-defined \emph{serialized representation}.
@@ -240,10 +244,10 @@ Such recovery process is called \emph{deserialization}.
 \label{sec:dsdl_bit_length_set}
 A serialized representation is a sequence of binary digits (bits);
 the number of bits in a serialized representation is called its \emph{bit length}.
-A \emph{bit length set} of a data type (or a data schema) refers to the set of bit length values of all possible
-serialized representations of objects that are instances of the data type (or that follow the data schema).
+A \emph{bit length set} of a data type refers to the set of bit length values of all possible
+serialized representations of objects that are instances of the data type.
 
-A data type or schema whose bit length set contains more than one element is said to be \emph{variable length}.
+A data type whose bit length set contains more than one element is said to be \emph{variable length}.
 The opposite case is referred to as \emph{fixed length}.
 
 The data type of a serialized message or service object exchanged over the network
@@ -251,3 +255,5 @@ is recovered from its subject-ID or service-ID, respectively,
 which is attached to the serialized object, along with other metadata, in a manner dictated by the applicable
 transport layer specification (chapter~\ref{sec:transport}).
 For more information on port identifiers and data type mapping refer to section~\ref{sec:basic_subjects_and_services}.
+
+The bit length set is not defined on service types (only on their inferiors: request and response).
--- a/specification/dsdl/architecture.tex
+++ b/specification/dsdl/architecture.tex
@@ -236,7 +236,9 @@ and they are composed of exactly two inner data type definitions\footnote{
 }.
 These inner types are the service request type and the service response type,
 separated by the \emph{service response marker}.
-They are otherwise ordinary data types except that they are unutterable\footnote{Cannot be referred to.}
+They are otherwise ordinary data types except that they are unutterable\footnote{%
+    Cannot be referred to. Another commonly used term is ``Voldemort type''.
+}
 and they derive some of their properties\footnote{Like version numbers or deprecation status.}
 from their \emph{parent service type}.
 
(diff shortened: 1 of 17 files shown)


specification/dsdl/architecture.tex
```diff
@@ -230,6 +228,18 @@ \subsection{Elements of data type definition}
 or assertions (special statements introduced by data type designers
 which are statically validated by DSDL processing tools).
 
+Service type definitions are a special case:
```
thirtytwobits, line 231: why not just say that service types are not serializable types? We'll just say they are a namespace with strict rules. We might even allow them to be different .dsdl files in the future; who knows?

specification/dsdl/compatibility.tex
```diff
@@ -143,6 +143,18 @@ \subsubsection{Versioning principles}
...
 
+All data types under the same name and major version number should share the same extent and the same finality status.
+It is therefore advised to:
+\begin{itemize}
+    \item Avoid marking types final, especially complex types,
+    because it is likely to render their evolution impossible.
+
+    \item When the first version is released, its extent should be sufficiently large
+    to permit addition of new fields in the future.
+    Since the value of extent does not affect the network traffic, it is safe to pick a large value
```
thirtytwobits, line 154: But this isn't true if you nest the type within another, correct? Let's not give people a false sense of freedom here. If you pad it out too much you may regret the decision later if you try to nest it or create an array of it.

specification/dsdl/serialization.tex
```diff
@@ -568,20 +562,166 @@ \subsubsection{Tagged unions}
...
+and the implicit zero extension (section \ref{sec:dsdl_serialization_implicit_zero_extension})
+rules apply.
+
+It is allowed for a final composite type to nest non-final composite types, and vice versa.
+No special rules apply in such cases.
+
+\begin{remark}
+    The resulting serialized representation of a delimited composite is identical to \verb|uint8[<2**32]|
+    (sans the higher alignment requirement).
+    The implicit array length prefix here is like the delimiter header,
```
thirtytwobits, line 623:
```suggestion
    The implicit array length field is like the delimiter header,
```

specification/dsdl/serialization.tex
```diff
@@ -568,20 +562,166 @@ \subsubsection{Tagged unions}
...
+
+\begin{remark}
+    Top-level objects do not require the delimiter header because the change in their length does not necessarily
+    affect the backward compatibility thanks to the implicit truncation rule
+    (section \ref{sec:dsdl_serialization_implicit_truncation}) and the implicit zero extension rule
+    (section \ref{sec:dsdl_serialization_implicit_zero_extension}).
+\end{remark}
+
+The delimiter header is an implicit field of type \verb|uint32| that encodes the length of the following
+serialized representation in bytes\footnote{%
```
thirtytwobits, line 607:
```suggestion
The delimiter header is an implicit field of type \verb|uint32| that encodes the length of the
serialized representation it preceeds in bytes\footnote{%
```

specification/dsdl/serializable_types.tex
```diff
@@ -281,51 +313,282 @@ \subsubsection{Dependencies}
...
+        }_{\substack{\text{Memory reserve} \\ \text{(none if final)}}}%
+    }^{\substack{%
+        \text{Extent} \\
+        \text{(memory requirement)}
+    }}
+    $$
+    \caption{Serialized representation and extent\label{fig:dsdl_extent}}
+\end{figure}
+
+The default extent of a non-final (delimited) definition is the following function
```
thirtytwobits, line 404: You're going to have to give the reader a spoiler at this point that there are an additional 32-bits hiding in here as the delimiter header. Without that called out explicitly you find yourself utterly confused at your assertion that the 64 bit structure in your example (for example) takes up 96 bits.

specification/dsdl/serialization.tex
thirtytwobits, line 607: Is this an apt description though? It really just encodes the offset in bytes that the next field starts at, right? The length of the datatype is still a property of the definition the transmitting node used and any variable-length array lengths.

specification/dsdl/serializable_types.tex
```diff
@@ -281,51 +313,282 @@ \subsubsection{Dependencies}
...
+    }.
+    A type that is not final is also referred to as \emph{delimited}.
+\end{description}
+
+The extent is the growth limit for the maximal bit length of the type as it evolves.
+The extent should be at least as large as the longest serialized representation of any compatible version of the type,
+which ensures that an agent leveraging a particular version can correctly process any other compatible
+version due to the avoidance of unintended truncation of serialized representations.
+
+Serialized representations of evolvable definitions carry additional metadata which introduces a certain overhead.
```
thirtytwobits, line 378: But only if they are nested, right? There isn't additional meta-data for top-level types.

thirtytwobits (changes requested):
We need to add a "changes since last revision" section and detail the structural changes between alpha and beta.

Let me make sure I understand what these are:

1. DSDL grammar: adding final keyword
2. DSDL grammar: adding extent keyword
3. all serializable types are now alignas(8)
4. extent types are always alignas(8) including primitive types (right?)
5. All types are automatically 32-bits bigger when nested inside of other types because there is an implicit 32-bit delimiter appended to them.

In general I think I'm okay with this but I'd like to have some of my questions answered (comments in this review) to make sure I'm not missing or misunderstanding anything.

(You are missing dsdl.tml from the clean script btw).

specification/dsdl/serializable_types.tex
```diff
@@ -195,20 +225,22 @@ \subsection{Array types}
...
-There are two kinds of composite data type definitions: message types and service types.
+There are two kinds of composite type definitions: message types and service types.
 All versions of a data type shall be of the same kind\footnote{%
     For example, if a data type version 0.1 is of a message kind, all later versions of it shall be messages, too.
 }.
 
-A message data type defines one data schema.
-
-A service data type contains two data schema definitions:
+A service type defines two unutterable inferior data types:
```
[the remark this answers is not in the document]
pavel-kirienko: They are "unutterable inferior types". An unutterable type (also known as [Voldemort type](https://wiki.dlang.org/Voldemort_types)) is a type that cannot be named. I think this is a fairly common slang term, do you think it is unacceptable here?

specification/dsdl/serializable_types.tex
```diff
@@ -195,20 +225,22 @@ \subsection{Array types}
...
 one for service request object, and one for service response object, in that order.
-The two schemas are separated by the service response marker (``\verb|---|'') on a separate line.
+The two inferiors are separated by the service response marker (``\verb|---|'') on a separate line.
```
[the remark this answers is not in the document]
pavel-kirienko: We already did:

https://github.com/UAVCAN/specification/blob/dc5be8f100bc7ba6b3c87cd557abbbcf303a0c17/specification/dsdl/grammar.parsimonious#L23

This parenthesized sample is not meant to override the syntax specification.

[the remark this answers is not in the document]
thirtytwobits: 👍

specification/dsdl/serializable_types.tex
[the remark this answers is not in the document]
thirtytwobits: Ah, no. I somehow missed this slang although I have heard of Voldemort types. I think putting those two words together makes for something that sound like a joke instead of a specification. If we can find a better term then "inferior" I think we could keep unutterable. Let's not have people refer to their carefully crafted data types as "a bunch of inferior crap so bad people can't even bring themselves to talk about it." :)

specification/dsdl/serializable_types.tex
```diff
@@ -281,51 +313,282 @@ \subsubsection{Dependencies}
...
+    It follows that empty composites can be inserted arbitrarily to force byte alignment
+    of the next field(s) at runtime.
+}.
+
+Per the definitions given in \ref{sec:dsdl_serializable_alignment_padding},
+a serialized representation of a composite type is padded to 8 bits by inserting padding bits
+after its last element until the resulting length is a multiple of 8 bits.
+
+Given a set of field attributes, their \emph{cumulative bit length set} is computed
+by evaluating every permutation of their respective bit length sets plus the required padding.
```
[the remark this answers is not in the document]
pavel-kirienko: Asymptotic complexity analysis assumes that the argument approaches the infinity, but in our case `n` is guaranteed to be small because data types contain few fields. I think the current performance issues in PyDSDL are rather caused by the implementation rather than specification. Even if it is not the case, I think it is the kind of problem that is safe to shelve until later.

specification/dsdl/serializable_types.tex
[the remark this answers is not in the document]
pavel-kirienko: Can you suggest a specific word?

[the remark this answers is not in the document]
thirtytwobits: Actually, it's more useful to start calling service "types" service namespaces or perhaps service tuples? Then these can remain datatypes.

specification/dsdl/architecture.tex
pavel-kirienko: Moving service types into a different category is a massive can of worms :canned_food: :snake: 

They have attributes that are inherent to serializable types like name, version, fixed port-ID, and deprecation status, so they are closer to being a special category than namespaces or regular serializable types. Do we want it changed in this PR? Should we postpone it until later? Do we keep it as-is? I tend to prefer the latter option right now.

thirtytwobits: Hmph. You are right. Okay, um, just call then service request and/or response types then. If we were to change how we defined service types this would remain the same terminology.

specification/dsdl/compatibility.tex
pavel-kirienko: >But this isn't true if you nest the type within another, correct?

Not correct, no. Extent does not exist on the wire. 

```python
# A.1.0
uint8 foo
@extent 1024
```

```python
# B.1.0
uint8 bar
A.1.0 a
uint8 baz
```

When B.1.0 is serialized with `foo` = 0x11, `bar` = 0x22, `baz` = 0x33, you get this on the wire:

```
22 01 00 00 00 11 33
^  \_________/  ^  ^
|   delimiter   |  |
|   header of   |  baz
|   "A.1.0 a"   |
bar             foo
```

However, to receive an object of type B.1.0, you will need to allocate a bit over 1 KiB of space to store its serialized representation because A.1.x can be up to 1 KiB large.

pavel-kirienko:
> all serializable types are now alignas(8)
> extent types are always alignas(8) including primitive types (right?)

Composite types are alignas(8). Primitives and voids are alignas(1) and an array of T is alignas(T).

> All types are automatically 32-bits bigger when nested inside of other types because there is an implicit 32-bit delimiter appended to them.

Not *all* types but only non-final composite types. Otherwise yes.

So if you mark your type `@final` (btw later I thought about saying "sealed" instead of "final" because it better describes the effect), you can nest it all you want and it wouldn't incur the 32-bit overhead.

thirtytwobits:
>  later I thought about saying "sealed" instead of "final" because it better describes the effect

Ooh. I like "sealed" better especially since final is a C++ keyword that has a very different effect.

specification/dsdl/serializable_types.tex
pavel-kirienko: The delimiter header is not involved here. The 96 bits are explained right under the example:

image

specification/dsdl/serialization.tex
pavel-kirienko: > It really just encodes the offset in bytes that the next field starts at, right?

Yes.

specification/dsdl/compatibility.tex
thirtytwobits: Okay. What I don't get is the assertion that you "need to allocate" the extent. Do you mean, "you need to be able to allocate at least the extent"? That is, if I do have dynamic memory there's nothing about this scheme that requires me to _always_ utilize the full extent worth to buffer a raw message nor deserialize a datatype regardless of the size on the wire, correct?

specification/dsdl/serializable_types.tex
```diff
@@ -1,15 +1,43 @@
...
+the resulting alignment or length requirements are satisfied.
+During deserialization, the padding bits are ignored (skipped) irrespective of their value
+(also see related section \ref{sec:dsdl_serialization_implicit_truncation}).
+
+\begin{remark}
+    For example, given a variable-length entity whose length varies between 1 and 3 bits,
+    followed by a field whose type has the alignment requirement of 8,
+    one may end up with 5, 6, or 7 bits of padding inserted before the second field at runtime.
+
+    The exact amount of such padding cannot always be determined statically because it depends on a runtime value;
```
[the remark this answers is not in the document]
thirtytwobits: Much better. Thanks

specification/dsdl/serialization.tex
```diff
@@ -568,20 +562,166 @@ \subsubsection{Tagged unions}
...
+        \item Encode $x_1 = 2$.
+        \item Memorize the current memory address $M_\text{current}$.
+        \item Go back to $M_\text{origin}$.
+        \item Encode a 32-bit wide value of $(M_\text{current} - M_\text{origin})$.
+        \item Go back to $M_\text{current}$.
+    \end{enumerate}
+
+    However, if the object is known to be of a constant size, the above can be simplified,
+    because there may be only one possible value of the delimiter header.
+    Automatic code generation tools should take advantage of this knowledge for best result.
```
thirtytwobits, line 726:
```suggestion
    Automatic code generation tools should take advantage of this knowledge.
```

thirtytwobits:
Okay, so I think the one sticking point for me is the default extent. The 1.5 value seems arbitrary. How did we arrive at this? Is there a precedence for a default memory reserve in other data interchange technologies?

specification/dsdl/serializable_types.tex
```diff
@@ -281,51 +314,290 @@ \subsubsection{Dependencies}
...
+        \text{Extent} \\
+        \text{(memory requirement)}
+    }}
+    $$
+    \caption{Serialized representation and extent\label{fig:dsdl_extent}}
+\end{figure}
+
+The default extent of a non-sealed (delimited) definition is the following function
+of the cumulative bit length set\footnote{Section \ref{sec:dsdl_composite_alignment_cumulative_bls}.}
+of its fields $B$:
```
thirtytwobits, line 407:
```suggestion
Where a non-sealed (delimited) type is defined \emph{without} an explicit extent value a default memory reserve of 50\% is implied using the following function
of the cumulative bit length set\footnote{Section \ref{sec:dsdl_composite_alignment_cumulative_bls}.}
of its fields $B$:
```

specification/dsdl/compatibility.tex
pavel-kirienko: Yes, correct. If we could rely on dynamic memory, we wouldn't need the extent at all. Cap'n'Proto allocates dynamic memory in fragments when it deserializes a message because the maximum footprint cannot be determined statically there.

specification/dsdl/serializable_types.tex
pavel-kirienko: > a default memory reserve of 50\% is implied using the following function

I don't like the redundancy here. The 50% overhead is defined in the equation already, why introduce it twice?

pavel-kirienko:
> Okay, so I think the one sticking point for me is the default extent. The 1.5 value seems arbitrary. How did we arrive at this? Is there a precedence for a default memory reserve in other data interchange technologies?

I couldn't locate similar practices anywhere, not even in ASN.1. It would appear that most standards have no problem with dynamic memory, which seems natural. The multiplier is, indeed, chosen arbitrarily, but it seems like a low-risk guess that can be revised easily if necessary.

> Regardless, adding in some of the discussion here: https://forum.uavcan.org/t/data-type-extensibility-and-composition/827/9 might help orient people about why we are wasting their precious memory.

Can you list any specific points that should be transferred? I was consulting with my own notes from that thread and I thought that I have carried over everything worthwhile but maybe I missed something.

> Also needed is a discussion of the effect of removing a field from a datatype that uses this default extent. By shrinking the serializable size of the type I'd also be shrinking the extent which seems problematic, right?

Yes, but I think this is covered adequately in 3.8.3.2, in the beginning here:

image

Do you think this needs to be expanded?

specification/dsdl/serializable_types.tex
thirtytwobits: I updated my suggestion.

thirtytwobits:
> > Regardless, adding in some of the discussion here: https://forum.uavcan.org/t/data-type-extensibility-and-composition/827/9 might help orient people about why we are wasting their precious memory.
> 
> Can you list any specific points that should be transferred? I was consulting with my own notes from that thread and I thought that I have carried over everything worthwhile but maybe I missed something.

Let's add a non-normative section that explains the 50% is arbitrarily chosen so nobody thinks that 2/3 is some magic number. So something like:

/begin{remark}
    Because of UAVCAN's commitment to determinism memory buffer allocation can become an issue. When using a flat encoding format with the implicit truncation rule, it is clear that the last defined fields are to be truncated out shall the buffer be too small to accommodate a received message. In a delimited (orthogonal) encoding, this behavior can no longer be relied on. See the rightmost figure:

image

Here, the field Y.c is truncated out even though one would expect the entire Y to be removed.

Conventional protocols like DDS manage this simply by delaying the memory requirement identification until runtime, which is unacceptable to UAVCAN. The solution for UAVCAN is to require implementations to reserve more memory than required by the data type definition unless it is @sealed (or unless the implementation does use dynamic memory allocation). The multiplier chosen to define the /emph{default} amount of reserve memory (i.e. the amount if a specific extent value is not provided) was chosen arbitrarily and is considered a common-sense default that is not optimized for any specific property of a system. This approach allows for limited extensibility while enabling deterministic memory requirements.
/end{remark}


> > Also needed is a discussion of the effect of removing a field from a datatype that uses this default extent. By shrinking the serializable size of the type I'd also be shrinking the extent which seems problematic, right?
>
> Do you think this needs to be expanded?

Perhaps we remind the user that, if they update a datatype by adding a field, they should also consider fixing the extent at the previously defined value.

pavel-kirienko:
I have accepted your suggestions but not as-is. I did not include the diagram from that forum thread because at the point where we introduce the extent, it is not known anything about delimited serialization, so we can't just show the delimiter header on a diagram without explaining what it is. Also, the diagram falls short of the quality bar set for the Specification. So, instead, I just added a link to an existing diagram at the end of the Serialization section, where I also added a few extra sentences about `@extent`.

thirtytwobits (approved):

thirtytwobits:
I'm approving. I will say for the record this change gives me some pause. Once you reason through it the rational becomes clear and it's a novel approach to force pre-allocating memory for future expansion on the system that is quite rational for real-time designs. My pause is that it is rather complicated by default which seems to go against the "U" in UAVCAN. I'm wondering if we are going to regret not making things sealed by default in the future more than we would have regretted not forcing users to consider extensibility by default.

pavel-kirienko:
Let's say, "U" is for "as simple as it can be, but not simpler"

PR #100, opened 2020-10-10
Be more explicit about the flexibility of the transfer-ID timeout setting

### LINKED ISSUE #78: Transfer-ID timeout auto-tuning
Consider mentioning in the Specification that the transfer-ID timeout can be auto-detected based on the incoming messaging frequency.

https://forum.uavcan.org/t/alternative-transport-protocols/324/19?u=pavel.kirienko

This change does not require modifying any of the existing specified behavior; rather, it's an implementation advice. Corner cases have to be analyzed first.

pavel-kirienko:
This was explicitly available in v0. The v1 Specification should speak more directly about this to make it clear that the scenario shown below is within the scope.

@JediJeremy in https://forum.uavcan.org/t/alternative-transport-protocols-in-uavcan/324/17:

> The CAN case has this covered, but If an arm takes a hit and you’re limping home with a broken wire and rebooting speed controller and have a serial-over-bluetooth redundant link, two seconds is a long, long time

Fixes #78

### DIFF
--- a/specification/transport/abstract.tex
+++ b/specification/transport/abstract.tex
@@ -669,9 +669,12 @@ the identity of the transport instance that carries the transfer\footnote{%
 \emph{Transfer reassembly} is the real-time process of reconstruction of the transfer payload and its metadata from
 a sequence of relevant transport frames.
 
-\emph{Transfer-ID timeout} is a time interval that is 2 (two) seconds long.
-The semantics of this entity are explained below.
-Implementations are allowed to redefine this value provided that such redefinition is explicitly documented.
+\emph{Transfer-ID timeout} is a time interval whose semantics are explained below.
+Implementations may define this value statically according to the application requirements.
+Implementations may automatically adjust this value per session at runtime as a function of the
+observed transfer reception interval.
+Transfer-ID timeout values greater than 2 (two) seconds are not recommended.
+Implementations should document the value of transfer-ID timeout or the rules of its computation.
 
 \emph{Transport frame reception timestamp} specifies the moment of time when the frame is received by a node.
 \emph{Transfer reception timestamp} is the reception timestamp of the earliest received frame of the transfer.
@@ -693,6 +696,11 @@ For a given session specifier, a successfully reassembled transfer that is tempo
 any other successfully reassembled transfer under the same session specifier by more than the transfer-ID timeout
 is considered unique regardless of its transfer-ID value.
 
+If the optimal transfer-ID timeout value for a given session cannot be known in advance,
+it can be computed at runtime as a multiple of the average transfer reception interval (on this session).
+The parameters of such computation are to be chosen according to the requirements of the application,
+but they should always be documented.
+
 \begin{remark}
     Low transfer-ID timeout values increase the risk of undetected transfer duplication when such transfers
     are significantly delayed due to network congestion,
@@ -700,6 +708,18 @@ is considered unique regardless of its transfer-ID value.
 
     High transfer-ID timeout values increase the risk of an undetected transfer loss
     when a remote node suffers a loss of state (e.g., due to a software reset).
+
+    The ability to auto-detect the optimal transfer-ID timeout value per session at runtime ensures that the
+    application can find the optimal balance even if the temporal properties of the network are not known in advance.
+    As a practical example, an implementation could compute the exponential moving average of the
+    transfer reception interval $x$ for a given session and define the transfer-ID timeout as $2x$.
+
+    It is important to note that the automatic adjustment of the transfer-ID timeout should only be done
+    on a per-session basis rather than for the entire port, because there may be multiple remote nodes
+    emitting transfers on the same port at different rates.
+    For example, if one node emits transfers at a rate $r$ transfers per second, and another node emits transfers
+    on the same port at a much higher rate $100r$, the resulting auto-detected transfer-ID timeout might be
+    too low, creating the risk of accepting duplicates.
 \end{remark}
 
 Implementations are recommended, but not required, to support reassembly of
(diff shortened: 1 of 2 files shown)


specification/transport/abstract.tex
```diff
@@ -662,16 +662,19 @@ \subsubsection{Transmission over redundant transports}
...
 a sequence of relevant transport frames.
 
-\emph{Transfer-ID timeout} is a time interval that is 2 (two) seconds long.
-The semantics of this entity are explained below.
-Implementations are allowed to redefine this value provided that such redefinition is explicitly documented.
+\emph{Transfer-ID timeout} is a time interval whose semantics are explained below.
+Implementations may define this value statically according to the application requirements.
+Implementations may automatically adjust this value per session at runtime as a function of the
+observed transfer reception interval.
+Transfer-ID timeout values greater than 2 (two) seconds are not recommended.
```
thirtytwobits, line 676: where do we specify what to do if a node gets a transfer using the same session identifier but with a new transfer id? My assumption is this should reset the pending session?

pavel-kirienko: Yes. In v0 we used to be extremely hands-on about the reception logic, but at one point I realized that it is too limiting because implementations should be able to choose specific approaches as long as the *externally observable behavior* is spec-compliant. In v1 we only state the high-level requirements and constraints, allowing implementers to decide how to meet them. To your question specifically, we say this:

**4.1.4.1 Definitions:** An ordered transfer sequence is a sequence of transfers whose temporal order is covariant with their transfer-ID values.

**4.1.4.2 Behaviors:** Reassembled transfers shall form an ordered transfer sequence.

Ergo, if a new transfer is started while the previous one is not yet completed, the node should discard the old transfer because not doing so in the hopes that it will be eventually completed after the new one is finished may violate the ordering requirement.

The specification is not a legal document though, and it is open to malicious interpretation. One could choose to do the following while being technically spec-conforming:

- Delay the reception of the new transfer indefinitely until the old one is completed.
- Drop the new transfer in favor of completing the old one.

I am not sure if we should bother explicitly banning such behaviors.

specification/transport/abstract.tex
```diff
@@ -693,13 +696,30 @@ \subsubsection{Behaviors}
...
     High transfer-ID timeout values increase the risk of an undetected transfer loss
     when a remote node suffers a loss of state (e.g., due to a software reset).
+
+    The ability to auto-detect the optimal transfer-ID timeout value per session at runtime ensures that the
+    application can find the optimal balance even if the temporal properties of the network are not known in advance.
+    As a practical example, an implementation could compute the exponential moving average of the
+    transfer reception interval $x$ for a given session and define the transfer-ID timeout as $2x$.
+
+    It is important to note that the automatic adjustment of the transfer-ID timeout should only be done
+    on a per-session basis rather than for the entire port, because there may be multiple remote nodes
```
thirtytwobits, line 718: To use a per-session average you need to start with a non-zero number so there's still some hard-coded default involved, correct?

specification/transport/abstract.tex
```diff
@@ -693,13 +696,30 @@ \subsubsection{Behaviors}
 any other successfully reassembled transfer under the same session specifier by more than the transfer-ID timeout
 is considered unique regardless of its transfer-ID value.
 
+If the optimal transfer-ID timeout value for a given session cannot be known in advance,
+it can be computed at runtime as a multiple of the average transfer reception interval (on this session).
```
thirtytwobits, line 700: Let's decline to comment on any specific algorithm in the normative part of the specification. I think there's more work one would need to do for this to be better then a hard-coded default especially given a per-session limit and for any system where most multi-part transfers are 2 or 3 frames.

specification/transport/abstract.tex
thirtytwobits: Dropping the old transfer and starting the new seems like the best light-weight behaviour. The other valid behaviour is to support multiple, interleaved transfers as long as all the transfer ID timeout never expires while each is being assembled. Of course this second, complex implementation only makes sense if a transmitting node has logic that may result in interleaved transfers (I suppose this becomes more plausible for nodes running a multi-tasking operating system that uses concurrent queues between the applications and the UAVCAN media layer). 

This is just me thinking this over. Not really a PR comment. I feel like I remember discussing this in the past but it's been awhile, sorry.

specification/transport/abstract.tex
thirtytwobits: I feel like we're also poking around at network congestion algorithms here. Is this something where I could implement a slow-start algorithm on the transmitter such that the receiver would get a slower but more consistent pacing of frames for a given transfer?

specification/transport/abstract.tex
pavel-kirienko: Moved into a footnote.

specification/transport/abstract.tex
pavel-kirienko: > To use a per-session average you need to start with a non-zero number so there's still some hard-coded default involved, correct?

Yes, but since we allow the application to override the transfer-ID, by extension we also allow it to choose an arbitrary initial value. That's how I intended it to read anyway.

> I feel like we're also poking around at network congestion algorithms here. Is this something where I could implement a slow-start algorithm on the transmitter such that the receiver would get a slower but more consistent pacing of frames for a given transfer?

You seem to be talking about traffic shaping (leaky bucket) logic here, but I don't think it is related to transfer-ID timeout. I expect the importance of traffic policy enforcement to decrease in a network that segregates traffic by priority levels (which is why it is rarely talked about in the context of CAN networks where there are 2^(29 or 11) priority levels). The reason being that if a high-priority class traffic unconditionally preempts low priority traffic, the latter can be excluded from analysis at this level. I expect that we will speak about network calculus at length when we get to switched networks, that is a separate topic. CAN is much simpler in this regard.

thirtytwobits (approved):

PR #103, opened 2020-10-31
Describe an edge case concerning deserialization of delimited objects

pavel-kirienko:
This is a minor change that eliminates the uncertainty that arises when the value of the delimiter header is X bytes whereas the serialized representation that contains it ends after N<X bytes.

### DIFF
--- a/specification/dsdl/serialization.tex
+++ b/specification/dsdl/serialization.tex
@@ -145,7 +145,8 @@ in the case of nested objects, by the \emph{delimiter header}
 
 \subsubsection{Error handling}\label{sec:dsdl_serialized_error}
 
-In this section and further, objects that nest other objects are referred to as \emph{outer objects}.
+In this section and further, an object that nests other objects is referred to as an \emph{outer object}
+in relation to the nested object.
 
 Correct UAVCAN types shall have no serialization error states.
 
--- a/specification/dsdl/serialization.tex
+++ b/specification/dsdl/serialization.tex
@@ -143,7 +143,9 @@ in the case of nested objects, by the \emph{delimiter header}
     \verb|array| contains a sequence of four zeros: $0, 0, 0, 0$.
 \end{remark}
 
-\subsubsection{Error handling}
+\subsubsection{Error handling}\label{sec:dsdl_serialized_error}
+
+In this section and further, objects that nest other objects are referred to as \emph{outer objects}.
 
 Correct UAVCAN types shall have no serialization error states.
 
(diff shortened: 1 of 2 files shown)


specification/dsdl/serialization.tex
```diff
@@ -143,7 +143,9 @@ \subsubsection{Implicit zero extension of missing data}\label{sec:dsdl_serializa
     \verb|array| contains a sequence of four zeros: $0, 0, 0, 0$.
 \end{remark}
 
-\subsubsection{Error handling}
+\subsubsection{Error handling}\label{sec:dsdl_serialized_error}
+
+In this section and further, objects that nest other objects are referred to as \emph{outer objects}.
```
thirtytwobits, line 148:
```suggestion
In this section and further, an object that nests other objects is referred to as an \emph{outer object}
in relation to the nested object. A \emph{top-level object} is an object that is not nested inside another but
which may be an outer object for types nested within it.
```

specification/dsdl/serialization.tex
```diff
@@ -614,6 +622,11 @@ \subsubsection{Delimited types}\label{sec:dsdl_serialization_composite_non_seale
 and the implicit zero extension (section \ref{sec:dsdl_serialization_implicit_zero_extension})
 rules apply.
 
+The length encoded in a delimiter header cannot exceed the number of bytes remaining between the delimiter header
+and the end of the serialized representation of the outer object.
```
thirtytwobits, line 626: Is this property transitive between all outer objects? i.e. should this read:
```suggestion
The length encoded in a delimiter header cannot exceed the number of bytes remaining between the delimiter header
and the end of the serialized representation of the top-level object.
```

thirtytwobits (changes requested):

specification/dsdl/serialization.tex
pavel-kirienko: This formulation implies that the length of the serialized representation of the next outer object in the hierarchy is irrelevant, which is incorrect. Suppose A (top-level object) nests B (outer object) which nests C. If the delimiter header of C specifies sizeof(C) bytes but the serialized representation of B ends in less than sizeof(C) bytes, the serialized representation of the entire structure (A,B, and C) is invalid. In your formulation, this would not be the case because it applies only to the top-level object A ignoring the direct container of C.

thirtytwobits: yeah. you're right.

thirtytwobits (approved):

PR #116, opened 2022-09-09
proposal for RFC process

thirtytwobits:


README.md
```diff
@@ -8,6 +8,27 @@ The sources of the Cyphal specification and other related documents are containe
 When cloning this repository, don't forget to initialize the Git submodules:
 `git submodule update --init --recursive`.
 
+## Governence
```
pavel-kirienko, line 11:
```suggestion
## Governance
```

README.md
```diff
@@ -8,6 +8,27 @@ The sources of the Cyphal specification and other related documents are containe
...
+
+The Cyphal specification is community-managed and completely open. Anyone can propose changes but only Cyphal maintainers may commit changes to this repository where the contents of the HEAD revision of the primary branch constitutes the latest version of the specification in-effect.
+
+### Request For Comments (RFC) Process
+
+Changes to the specification shall use the following, community-driven RFC process:
+
+1. An RFC is posted on the [OpenCyphal forum's specification topic](https://forum.opencyphal.org/c/dev/spec/15) with the details of the proposed change. The title of this post must start with "RFC: " to indicate it is the start of a proposed change to the specification.
```
pavel-kirienko, line 19:
```suggestion
1. An RFC is posted on the [OpenCyphal forum's specification category](https://forum.opencyphal.org/c/dev/spec/15) with the details of the proposed change. The title of this post should start with "RFC: " to indicate it is the start of a proposed change to the specification.
```

README.md
```diff
@@ -8,6 +8,27 @@ The sources of the Cyphal specification and other related documents are containe
...
+   - An RFC shall remain open for comments for at least 10 days from the date it was sponsored regardless of the sponsoring Maintainer's opinion.
+   - Any maintainer can remove any open RFC at any time, without regard to any other rule, if it or the subsequent discussion violates the forum's community standards.
+   - An open RFC can be removed from the forum, without regard to any other rule, if two or more maintainers agree to remove it.
+4. Once an RFC is ready for a final disposition at least one maintainer, _other than the sponsoring maintainer_, and the majority of active maintainers shall determine if the RFC is accepted or rejected.
+    - Rejected RFCs can be resubmitted after a 3-month cooling-off period.
+5. After an RFC is accepted on the forum the change can be made to the specification in this repository. The Pull Request must reference the forum discussion and at least one maintainer must agree that the change is correct as specified by the text of the approved RFC.
+    - If an RFC is not reified by a change to the Specification within three months of approval then it expires. Expired RFCs can be resubmitted after a 3-month cooling off period.
```
pavel-kirienko, line 30: This expiration clause effectively means that if a proposal is not implemented in 3 months then one would have to wait extra three months before it is reintroduced.

pavel-kirienko (commented):
I approve the direction in which this is going but at the same time it feels like imposing too much structure upon our lax workflow. Is there an option to reduce the text by about three-fold to let it state only the following general ideas: RFCs are proposed on the forum, after some initial discussion they are promoted to pull requests, and if the maintainers are happy, the pull requests are accepted; the maintainers undertake to act in the best interest of the project? I don't think we have a sufficient track record of accepting external RFCs to justify the existence of strict regulation here.

README.md
thirtytwobits: Well, by "implemented" we mean the specification is updated, right?

I'm trying avoid having a backlog of approved RFCs which would mean you need to know the current specification + the set of approved but not implemented RFCs to know the current state of the standard. My reasoning is that if the community is indifferent but willing to approve an RFC all the same _someone_ has to want it bad enough to make the change. If nobody does after the three months then nobody really cared enough and it probably should die.

The three-month thing in all of these steps is to avoid malcontents from abusing the RFC process to badger the community into accepting a change they repeatedly reject.

I'm open to modifying this, however. What would you propose?

thirtytwobits:
> I approve the direction in which this is going but at the same time it feels like imposing too much structure upon our lax workflow. Is there an option to reduce the text by about three-fold to let it state only the following general ideas: RFCs are proposed on the forum, after some initial discussion they are promoted to pull requests, and if the maintainers are happy, the pull requests are accepted; the maintainers undertake to act in the best interest of the project? I don't think we have a sufficient track record of accepting external RFCs to justify the existence of strict regulation here.

I'm open to de-lawyer this text by about half but I also want to communicate that maintainers have a responsibility to not just ram things through. There has to be some amount of time an RFC is left open for comment. How about something more like, RFCs are proposed on the forum, after some initial discussion and at least 10 days, they are promoted to pull requests, and if the maintainers are happy, the pull requests are accepted; the maintainers undertake to act in the best interest of the project?

pavel-kirienko:
That sounds agreeable.

README.md
pavel-kirienko: I would replace the three-month period with a "reasonable timeframe".

README.md
```diff
@@ -16,18 +16,10 @@ The Cyphal specification is community-managed and completely open. Anyone can pr
...
-   - An open RFC can be removed from the forum, without regard to any other rule, if two or more maintainers agree to remove it.
-4. Once an RFC is ready for a final disposition at least one maintainer, _other than the sponsoring maintainer_, and the majority of active maintainers shall determine if the RFC is accepted or rejected.
-    - Rejected RFCs can be resubmitted after a 3-month cooling-off period.
-5. After an RFC is accepted on the forum the change can be made to the specification in this repository. The Pull Request must reference the forum discussion and at least one maintainer must agree that the change is correct as specified by the text of the approved RFC.
-    - If an RFC is not reified by a change to the Specification within three months of approval then it expires. Expired RFCs can be resubmitted after a 3-month cooling off period.
+1. An RFC is posted on the [OpenCyphal forum's specification topic](https://forum.opencyphal.org/c/dev/spec/15) with the details of the proposed change. The title of this post should start with "RFC: " to indicate it is the start of a proposed change to the specification.
```
pavel-kirienko, line 19: Not sure if this is entirely acceptable but the original wording seems incorrect.

```suggestion
1. An RFC is posted on the [OpenCyphal forum's specification section](https://forum.opencyphal.org/c/dev/spec/15) with the details of the proposed change. The title of this post should start with "RFC: " to indicate it is the start of a proposed change to the specification.
```

pavel-kirienko (dismissed):
LGTM, minimalistic and to the point.

pavel-kirienko (approved):

SchoberMJ (commented):
Looks good to me. 

How do we go from a specification pull-request to a functional change in the OpenCyphal repositories? Will the RFC process cover required changes to existing libraries or will that be left as tribal knowledge for the maintainers to manage behind the scenes?

thirtytwobits:
> How do we go from a specification pull-request to a functional change in the OpenCyphal repositories? 

This is not detailed as we are trying to keep this process informal, however, my expectation is the blast-radius of a given change would be a key discussion on the forum review of the RFC. This would include concerns by the community about their own implementations as well as maintainers considering how the proposed change would be rolled out in the reference implementations.

PR #126, opened 2023-04-22
Explain the portmanteau and reserve special comment forms

### LINKED ISSUE #114: Explain Cyphal portmanteau in Specfication
Update the line in the Specification which says:

> The name Cyphal stands for Uncomplicated Application-level Vehicular Computing And Networking.

To explain the portmanteau and mentions the historical transition.

### LINKED ISSUE #125: Reserve the special comment form
Reserve comments following `^#\[.+\]` for vendor extensions. See https://github.com/samcrow/canadensis/issues/16

pavel-kirienko:
Fix #114 
Fix #125

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -55,5 +55,6 @@
     ],
     "spellright.language": [
         "en_US"
-    ]
+    ],
+    "cSpell.enabled": false,
 }
--- a/specification/dsdl/grammar.tex
+++ b/specification/dsdl/grammar.tex
@@ -253,3 +253,12 @@ in this version of the DSDL specification, but this may change in the future.
     \texttt{lpt\textbackslash{}d}                      & \texttt{lpt9}      & Compatibility with Microsoft Windows \\
     \texttt{\_.*\_}                                    & \texttt{\_offset\_}& Special-purpose intrinsic entities \\
 \end{CyphalSimpleTable}
+
+\subsection{Reserved comment forms}\label{sec:dsdl_reserved_comment}
+
+Line comments that match the following regular expression are reserved for vendor-specific language extensions:
+\verb|^#\[.+\].*|
+
+\begin{remark}
+    For example, \verb|#[canadensis(enum_start)]|.
+\end{remark}
--- a/.vscode/spellright.dict
+++ b/.vscode/spellright.dict
@@ -482,3 +482,4 @@ cyber
 hyphal
 symbionts
 mycorrhizal
+canadensis
--- a/.vscode/spellright.dict
+++ b/.vscode/spellright.dict
@@ -476,3 +476,5 @@ BLS
 evolvable
 Protobuf
 deserialize
+Cyphal
+opencyphal
--- a/.vscode/spellright.dict
+++ b/.vscode/spellright.dict
@@ -478,3 +478,7 @@ Protobuf
 deserialize
 Cyphal
 opencyphal
+cyber
+hyphal
+symbionts
+mycorrhizal
--- a/specification/introduction/introduction.tex
+++ b/specification/introduction/introduction.tex
@@ -15,7 +15,18 @@ Cyphal can be approximated as a highly deterministic decentralized object reques
 with a specialized interface description language and a highly efficient data serialization format
 suitable for use in real-time safety-critical systems with optional modular redundancy.
 
-The name Cyphal stands for \emph{Uncomplicated Application-level Vehicular Computing And Networking}.
+``Cyphal'' is an invented word; a portmanteau of ``cyber'' and ``hyphal''.
+The former references cyber-physical systems,
+which is a generalization of the type of system this new protocol is optimized for.
+The latter describes hypha --- branching structures found in the fungal symbionts of mycorrhizal networks\footnote{%
+    A mycorrhizal network is an underground network found in forests and other plant communities,
+    created by the hyphae of mycorrhizal fungi joining with plant roots.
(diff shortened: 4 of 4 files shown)


specification/dsdl/grammar.tex
```diff
@@ -253,3 +253,12 @@ \subsection{Reserved identifiers}\label{sec:dsdl_reserved_identifiers}
...
     \texttt{\_.*\_}                                    & \texttt{\_offset\_}& Special-purpose intrinsic entities \\
 \end{CyphalSimpleTable}
+
+\subsection{Reserved comment forms}\label{sec:dsdl_reserved_comment}
+
+Line comments that match the following regular expression are reserved for vendor-specific language extensions:
+\verb|^#\[.+\].*|
+
+\begin{remark}
+    For example, \verb|#[canadensis(enum_start)]|.
```
thirtytwobits, line 263: Shouldn't this be minted? As-is, the period is ambiguous. Perhaps:

```
    \begin{minted}{python}
        #[canadensis(enum_start)] all of this line is part of the reserved form.
        # After the newline this comment is now a standard DSDL comment.
    \end{minted}
```

thirtytwobits (changes requested):

specification/dsdl/grammar.tex
pavel-kirienko: Updated as you suggested.

.vscode/settings.json
```diff
@@ -55,5 +55,6 @@
     ],
     "spellright.language": [
         "en_US"
-    ]
-}
\ No newline at end of file
+    ],
+    "cSpell.enabled": false,
```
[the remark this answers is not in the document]
pavel-kirienko: done

pavel-kirienko:
Things broke unexpectedly when I attempted to switch to the newer container. Please ignore the fact that the CI is red, this is not related to the main changes in this PR. I am going to fix the build later.

.vscode/settings.json
```diff
@@ -44,16 +44,356 @@
         }
     ],
     "latex-workshop.viewer.pdf.internal.port": 41491,
-    "spellright.documentTypes": [
-        "markdown",
-        "latex",
-        "plaintext"
+    "spellright.documentTypes": [ ],
```
thirtytwobits, line 47: missed one

specification/dsdl/grammar.tex
```diff
@@ -253,3 +253,15 @@ \subsection{Reserved identifiers}\label{sec:dsdl_reserved_identifiers}
     \texttt{lpt\textbackslash{}d}                      & \texttt{lpt9}      & Compatibility with Microsoft Windows \\
     \texttt{\_.*\_}                                    & \texttt{\_offset\_}& Special-purpose intrinsic entities \\
 \end{CyphalSimpleTable}
+
+\subsection{Reserved comment forms}\label{sec:dsdl_reserved_comment}
+
+Line comments that match the following regular expression are reserved for vendor-specific language extensions:
+\verb|^#\[.+\].*|
```
thirtytwobits, line 260: shouldn't this be `^\s*#\s*\[.+\].*`? Either way, it seems like a complex pattern to reserve. What about just `^\s*#!.*` so:

```
# This is a commen[sic]
#! This is an extension. All if it. The entire line!
# Back to business.

```

.vscode/settings.json
pavel-kirienko: This one is left on purpose to avoid having to manually disable the spellright extension if it's installed. I couldn't find a better way to disable it via the config files.

specification/dsdl/grammar.tex
pavel-kirienko: @samcrow could you please provide your input on this?

specification/dsdl/grammar.tex
```diff
@@ -253,3 +253,15 @@ \subsection{Reserved identifiers}\label{sec:dsdl_reserved_identifiers}
...
+
+\begin{remark}
+    \begin{minted}{python}
+        #[canadensis(enum_start)] all of this line is part of the reserved form.
```
samcrow, line 264: Minor note: Someone suggested `#[canadensis(enum_start)]`, but the directive I actually implemented is `#[canadensis(enum)]`.
To avoid confusion, I think it would be better to use `#[canadensis(enum)]`, or make up a different name for the example.

specification/dsdl/grammar.tex
pavel-kirienko: While waiting for Sam to respond, I want to add that this pattern is based on Sam's work on Canadensis, which derives this syntax from Rust, which in turn borrowed it from C#, so we appear to be in good company. There are no strong reasons to maintain consistency with these languages, nor there appear to be reasons to deviate.

samcrow: These are the thoughts I have right now:
* The current ideas use a # followed by a key character, which could be [, !, or something else.
* Allowing whitespace before the # seems like a good idea to be consistent with other languages I've used.
* Allowing whitespace between the # and the key character seems like not a good idea because it would allow breaking up the special sequence of characters.
* The choice of the key character doesn't really matter unless it's something commonly used to start sentences.

specification/dsdl/grammar.tex
pavel-kirienko: updated

pavel-kirienko:
eh perhaps but filtering the words manually doesn't sound like a good way to spend time.

PR #127, opened 2023-05-01
Specify Cyphal/UDP [experimental]

### LINKED ISSUE #98: Specify Cyphal/UDP
The draft is available at https://pycyphal.readthedocs.io/en/stable/api/pycyphal.transport.udp.html

The discussion is at https://forum.opencyphal.org/t/alternative-transport-protocols-in-uavcan/324

pavel-kirienko:
- Add Cyphal/UDP specification (marked as experimental); fixes #98.
- Move CRC implementations to the appendix.

The specification is quite concise because the key concepts are already covered in section "4.1 Abstract concepts". In the new section, I only had to cover the layouts.


maksimdrachov:
To save everyone a bit of time:

[Cyphal_Specification.pdf](https://github.com/OpenCyphal/specification/files/11372849/Cyphal_Specification.pdf)

maksimdrachov:
image

I think: `SUBJECT_ID_MASK = 2**15 - 1`?

specification/appendices/crc.tex
```diff
@@ -0,0 +1,242 @@
...
+
+This algorithm is also known as CRC-16/AUTOSAR or CRC-16/IBM-3740.
+Not to be confused with CRC-16/KERMIT.
+
+Width 16 bits,
+polynomial $\mathrm{1021}_{16}$,
+initial value $\mathrm{FFFF}_{16}$,
+not reflected,
+no output XOR.
+The native byte order is big endian.
```
maksimdrachov, line 13: Maybe make this a bulleted list?

```
This algorithm has the following properties:

\begin{itemize}
  \item Width: 16 bits
  \item Polynomial: $\mathrm{1021}_{16}$,
  \item Initial value: $\mathrm{FFFF}_{16}$,
  \item Not reflected
  \item No output XOR
  \item Native byte order is big endian
\end{itemize}
```

As it stands, it looks like this in the pdf:

image

(same remark for CRC-32C)

specification/appendices/crc.tex
```diff
@@ -0,0 +1,242 @@
...
+
+The value for an input sequence $\left(49, 50, \ldots, 56, 57\right)$ is $\mathrm{29B1}_{16}$.
+
```
maksimdrachov, line 16: Tested using following script:

<details>
<summary>Test script</summary>

```py
class CRC16CCITT:
    def __init__(self) -> None:
        assert len(self._TABLE) == 256
        self._value = 0xFFFF

    def add(self, data: bytes | bytearray | memoryview) -> None:
        val = self._value
        for x in data:
            val = ((val << 8) & 0xFFFF) ^ self._TABLE[(val >> 8) ^ x]
        self._value = val

    def check_residue(self) -> bool:
        return self._value == 0

    @property
    def value(self) -> int:
        return self._value

    @property
    def value_as_bytes(self) -> bytes:
        return self.value.to_bytes(2, "big")

    # fmt: off
    _TABLE = [
        0x0000, 0x1021, 0x2042, 0x3063, 0x4084, 0x50A5, 0x60C6, 0x70E7,
        0x8108, 0x9129, 0xA14A, 0xB16B, 0xC18C, 0xD1AD, 0xE1CE, 0xF1EF,
        0x1231, 0x0210, 0x3273, 0x2252, 0x52B5, 0x4294, 0x72F7, 0x62D6,
        0x9339, 0x8318, 0xB37B, 0xA35A, 0xD3BD, 0xC39C, 0xF3FF, 0xE3DE,
        0x2462, 0x3443, 0x0420, 0x1401, 0x64E6, 0x74C7, 0x44A4, 0x5485,
        0xA56A, 0xB54B, 0x8528, 0x9509, 0xE5EE, 0xF5CF, 0xC5AC, 0xD58D,
        0x3653, 0x2672, 0x1611, 0x0630, 0x76D7, 0x66F6, 0x5695, 0x46B4,
        0xB75B, 0xA77A, 0x9719, 0x8738, 0xF7DF, 0xE7FE, 0xD79D, 0xC7BC,
        0x48C4, 0x58E5, 0x6886, 0x78A7, 0x0840, 0x1861, 0x2802, 0x3823,
        0xC9CC, 0xD9ED, 0xE98E, 0xF9AF, 0x8948, 0x9969, 0xA90A, 0xB92B,
        0x5AF5, 0x4AD4, 0x7AB7, 0x6A96, 0x1A71, 0x0A50, 0x3A33, 0x2A12,
        0xDBFD, 0xCBDC, 0xFBBF, 0xEB9E, 0x9B79, 0x8B58, 0xBB3B, 0xAB1A,
        0x6CA6, 0x7C87, 0x4CE4, 0x5CC5, 0x2C22, 0x3C03, 0x0C60, 0x1C41,
        0xEDAE, 0xFD8F, 0xCDEC, 0xDDCD, 0xAD2A, 0xBD0B, 0x8D68, 0x9D49,
        0x7E97, 0x6EB6, 0x5ED5, 0x4EF4, 0x3E13, 0x2E32, 0x1E51, 0x0E70,
        0xFF9F, 0xEFBE, 0xDFDD, 0xCFFC, 0xBF1B, 0xAF3A, 0x9F59, 0x8F78,
        0x9188, 0x81A9, 0xB1CA, 0xA1EB, 0xD10C, 0xC12D, 0xF14E, 0xE16F,
        0x1080, 0x00A1, 0x30C2, 0x20E3, 0x5004, 0x4025, 0x7046, 0x6067,
        0x83B9, 0x9398, 0xA3FB, 0xB3DA, 0xC33D, 0xD31C, 0xE37F, 0xF35E,
        0x02B1, 0x1290, 0x22F3, 0x32D2, 0x4235, 0x5214, 0x6277, 0x7256,
        0xB5EA, 0xA5CB, 0x95A8, 0x8589, 0xF56E, 0xE54F, 0xD52C, 0xC50D,
        0x34E2, 0x24C3, 0x14A0, 0x0481, 0x7466, 0x6447, 0x5424, 0x4405,
        0xA7DB, 0xB7FA, 0x8799, 0x97B8, 0xE75F, 0xF77E, 0xC71D, 0xD73C,
        0x26D3, 0x36F2, 0x0691, 0x16B0, 0x6657, 0x7676, 0x4615, 0x5634,
        0xD94C, 0xC96D, 0xF90E, 0xE92F, 0x99C8, 0x89E9, 0xB98A, 0xA9AB,
        0x5844, 0x4865, 0x7806, 0x6827, 0x18C0, 0x08E1, 0x3882, 0x28A3,
        0xCB7D, 0xDB5C, 0xEB3F, 0xFB1E, 0x8BF9, 0x9BD8, 0xABBB, 0xBB9A,
        0x4A75, 0x5A54, 0x6A37, 0x7A16, 0x0AF1, 0x1AD0, 0x2AB3, 0x3A92,
        0xFD2E, 0xED0F, 0xDD6C, 0xCD4D, 0xBDAA, 0xAD8B, 0x9DE8, 0x8DC9,
        0x7C26, 0x6C07, 0x5C64, 0x4C45, 0x3CA2, 0x2C83, 0x1CE0, 0x0CC1,
        0xEF1F, 0xFF3E, 0xCF5D, 0xDF7C, 0xAF9B, 0xBFBA, 0x8FD9, 0x9FF8,
        0x6E17, 0x7E36, 0x4E55, 0x5E74, 0x2E93, 0x3EB2, 0x0ED1, 0x1EF0,
    ]


input_sequence = list(range(49, 58))

print(input_sequence)

crc = CRC16CCITT()

for i in input_sequence:
    i_in_bytes = i.to_bytes(1, byteorder="big")
    crc.add(i_in_bytes)

print(crc.value)
print(hex(crc.value))
```
</details>

Result is correct:

image

Same test for CRC32:

image

Interestingly, in the test script I've tried to change byte order for the input:

```py
for i in input_sequence:
    i_in_bytes = i.to_bytes(1, byteorder="little") # changing this to big has no effect on end result
    crc.add(i_in_bytes)
```

Assuming this is normal?

pavel-kirienko:
>I think: SUBJECT_ID_MASK = 2**15 - 1?

@maksimdrachov Currently, valid subject-ID values are from 0 to 8191 inclusive:

image

This range may be increased in the future. Both Cyphal/UDP and Cyphal/CAN are ready for that by providing extra bits for the identifiers that are currently unused.

specification/transport/udp/udp.tex
```diff
@@ -0,0 +1,337 @@
...
+uint4 version
+# The version of the header format. This document specifies version 1.
+# Packets with an unknown version number must be ignored.
+
+void4
+
+uint3 priority
+# The values are assigned from 0 (HIGHEST priority) to 7 (LOWEST priority).
+# The numerical priority identifiers are chosen to be consistent with Cyphal/CAN.
+# Notice that this is the opposite of the priority ordering in the IP DSCP field.
```
thirtytwobits, line 251: see comments above and revise accordingly.

specification/transport/udp/udp.tex
```diff
@@ -0,0 +1,337 @@
...
+    the packet if the \verb|end_of_transfer| bit is set,
+    unless the transfer CRC field has spilled over to the next frame
+    (in which case the frame would contain only the transfer CRC itself or the tail thereof).
+\end{remark}
+
+\subsection{Maximum transmission unit}
+
+In this section, the maximum transmission unit (MTU) is defined as the maximum size of a UDP/IP datagram payload.
+
+This specification does not restrict the MTU of the underlying transport.
```
thirtytwobits, line 333: We should take a harder stance here. A minimum MTU should be enforced that allows all regulated messages to be transfered in single frame.

specification/transport/udp/udp.tex
```diff
@@ -0,0 +1,337 @@
...
+It is recommended, however, to avoid MTU values less than 508~bytes to allow reduced (simplified) implementations
```
thirtytwobits, line 334: How about:

By setting MTU values greater than or equal to 508 bytes, any system that does not define messages larger than 480 bytes ($508 - 24 - 4$) can simplify their implementation by omitting support for multi-frame transfers.

specification/transport/udp/udp.tex
```diff
@@ -0,0 +1,337 @@
...
+of the protocol that do not require large transfer payloads to omit support for multi-frame transfers.
```
thirtytwobits, line 335: What about IP fragmentation and reassembly? Do we require and/or recommend that IP framing be used in lieu of Cyphal framing? Do we prohibit it? It seems complicated, if hand-writing a Cyphal/UDP/IP stack, to have to support both.

thirtytwobits (changes requested):

specification/transport/udp/udp.tex
```diff
@@ -0,0 +1,337 @@
...
+    still needs to obtain a node-ID value to participate in a Cyphal/UDP network.
+    This may be done either through manual assignment or by using the plug-and-play node-ID allocation service
+    (section~\ref{sec:application_functions_pnp}).
+}.
+The set of valid node-ID values for Cyphal/UDP is $[0, 65534]$.
+Value 65535 is reserved to represent both the broadcast and anonymous node-ID, depending on context.
+
+Sources of Cyphal/UDP traffic should set the packet TTL to 16 or higher\footnote{%
+    RFC~1112 prescribes a default TTL of 1,
+    but this is not sufficient for most applications Cyphal/UDP is intended for.
```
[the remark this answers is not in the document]
maksimdrachov: Pavel:

image

specification/transport/udp/udp.tex
```diff
@@ -0,0 +1,337 @@
...
+
+This section specifies a concrete transport based on the UDP/IPv4 protocol\footnote{%
+    Support for IPv6 may appear in future versions of this specification.
+}, as specified in IETF RFC~768.
+\textbf{
+    As of this version, the Cyphal/UDP specification remains experimental.
+    Breaking changes affecting wire compatibility are possible.
+}
+
+Cyphal/UDP is intended for low-latency, high-throughput intravehicular Ethernet networks with complex topologies,
```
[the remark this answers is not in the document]
pavel-kirienko: It's possible, but wouldn't it sound a bit defensive? The fact that it is not an encapsulation becomes apparent once you're done reading the first two pages.

.vscode/settings.json
```diff
@@ -63,8 +63,10 @@
         "attrib",
         "Autoconfigured",
         "autodetect",
+        "AUTOSAR",
```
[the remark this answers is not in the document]
pavel-kirienko: Can you point me to the PDF warnings in the CI output? I couldn't find them there. Although, either way I would be inclined to address them separately later.

The other two (hypcap and the uncyphalized link) are addressed.

Aside from these issues, there is another one that will need fixing one day: we rely on `tabu` for tables, which is no longer maintained and has many issues. At one point we should switch either to `tabularray` or the standard `tabular` environment (they both have strengths and drawbacks). Related: https://github.com/Zubax/document_templates/commit/b50b4e0e36d58bad27a701adb64aaa24cd07a78a

specification/transport/udp/udp.tex
```diff
@@ -0,0 +1,337 @@
...
+should be populated based on the Cyphal transfer priority level (section~\ref{sec:transport_transfer_priority})
+such that the maximum Cyphal priority level corresponds to class selector CS7
+and the minimum Cyphal priority level corresponds to class selector CS0,
+with the intermediate values mapped following the same principle.
+
+\begin{remark}
+    \begin{CyphalSimpleTable}{
+        Recommended DSCP class selector values\label{table:transport_udp_priority}
+    }{|l l l l|}
+        Cyphal priority & DSCP class selector & IP precedence value & IP precedence name      \\
```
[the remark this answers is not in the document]
pavel-kirienko: The reason I went with my original suggestion is the expectation that intra-vehicular routing policies will likely be configured from scratch and may not inherit the policies suggested in certain RFCs. However, this is not to say that consistency with COTS has no value.

The general principle behind the original IP precedence field is that higher values correspond to higher priorities (RFC791), which is upheld for reasons of backward compatibility in the DSCP field (RFC2474). Then, RFC3662 introduces the LE policy with the suggestion to mark it as DSCP CS1, but unless I am missing something, it makes no statement that CS1 should be treated as a lower-priority class compared to CS0, and indeed that would violate RFC791 and RFC1347 (the relevant part is *"If a host makes appropriate use of the TOS facility, its network service should be at least as good as (and hopefully better than) it would have been if the host had not used the facility"*). Therefore the DSCP values you suggested for the Low and Optional priority levels appear incorrect. Would you agree?

If we agree that the general principle of higher precedence corresponding to higher priority is likely to be upheld in all networks we care about, we could just as well keep the simple original mapping I initially proposed. Otherwise, we need to revise the lower levels.

specification/transport/udp/udp.tex
pavel-kirienko: I understand that you are suggesting inverting the priority mapping here, such that higher values correspond to higher priority. I think retaining consistency with Cyphal/CAN and the existing experimental implementations of Cyphal/UDP is desirable, if only for the sake of consistency itself, so I suggest keeping this unchanged.

specification/transport/udp/udp.tex
pavel-kirienko: The regulated messages have no upper limit on the maximum extent. We could define the lower MTU limit such that it is large enough to contain the largest regulated message available today, but this would not allow us to guarantee that it will be sufficient for possible future revisions of the regulated messages. Hence I don't think it's practical to tie the minimum MTU definition to DSDL.

specification/transport/udp/udp.tex
pavel-kirienko: I am inclined to prohibit multipart IP transfers as they are indeed redundant in the presence of Cyphal framing, but wouldn't that be overreach on our part? We define the protocol in terms of UDP/IP, and, strictly speaking, most of the things that happen underneath are not our responsibility.

.vscode/settings.json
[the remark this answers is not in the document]
thirtytwobits: I'm seeing a few of these when building with the container:

```
pdfTeX warning: pdflatex (file ./application/NED_ECEF-eps-converted-to.pdf): PDF inclusion: found PDF version <1.7>, but at most version <1.5> allowed
```

specification/transport/udp/udp.tex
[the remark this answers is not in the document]
thirtytwobits: My thinking is just to break a confirmation bias where a reader is starting on this section expecting to find out how to encapsulate Cyphal/CAN over UDP (i.e. the meta-transport types). Rather than defensive, I'd see it as a kindness to avoid confusion in such cases. Something like "dear reader - if you want to encapsulate CAN over UDP you can use meta-transport.can over Cyphal/UDP, which is a first-class IP protocol..." etc.

specification/transport/udp/udp.tex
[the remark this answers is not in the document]
thirtytwobits: I think we should bias towards enabling no-config behavioiur that is optimal for Cyphal/UDP and develop Cyphal/TSN as the protocol one should use when tuning a vehicle network to be optimal and deterministic. For UDP we should look to switches like [this](https://www.amazon.com/Ethernet-Splitter-Optimization-Unmanaged-TL-SG105/dp/B00A128S24) where IP telephony is pre-configured. So, actually, let me modify my proposal to exclude the AF41 and just support three types: CS1/0/lowest == optional (it does seem that this is actually lower than 1), DF/0/default == low, and EF/46/High == nominal.

Finally, note that value 7 is actually reserved for future use and may not be routed on some networks so a flat mapping is both wrong (i.e. 1 and 0 may be inverted in priority) and illegal (in that 7 is reserved).

.vscode/settings.json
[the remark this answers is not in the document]
pavel-kirienko: Please make sure you're using the same container version as the CI. If you are and the issue persists, let us please move it to a new ticket. I am going to mark this one resolved.

specification/transport/udp/udp.tex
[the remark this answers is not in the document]
pavel-kirienko: I have updated this part as discussed on Matrix: https://github.com/OpenCyphal/specification/pull/127/commits/1c33eaf32d03a769cec02bdc066f4e2b40a588dc. The specific level mapping may require bikeshedding but we need to keep in mind that per RFC8837, "excess" EF traffic is dropped. We may want to consider replacing EF with AF41.

specification/transport/udp/udp.tex
thirtytwobits: This continues to vex me. It's important, for the smallest and/or most critical nodes, that the amount of software required to run Cyphal is minimised. This means the network designer has to chose between ip framing or Cyphal. I'd like to take some stance on this and I'm leaning towards prohibiting multi-part Cyphal messages on IP networks. What reason do we have to go the other way and prohibit ethernet framing in favor of Cyphal?

specification/transport/udp/udp.tex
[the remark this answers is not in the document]
pavel-kirienko: Indeed, it is hard to come up with a sensible recommendation given the uncertainty. I will update this to require zero by default with configurability.

specification/transport/udp/udp.tex
```diff
@@ -0,0 +1,370 @@
...
+    (in which case the frame would contain only the transfer CRC itself or the tail thereof).
+\end{remark}
+
+\subsection{Maximum transmission unit}
+
+In this section, the maximum transmission unit (MTU) is defined as the maximum size of a UDP/IP datagram payload.
+
+This specification does not restrict the MTU of the underlying transport.
+It is recommended, however, to avoid MTU values less than 508~bytes,
+allowing applications to exchange up to $508 - 24 - 4 = 480$ bytes of payload in a single-frame transfer.
```
thirtytwobits, line 357: Should we recommend 1280 octets since that's the IPV6 link minimum (per https://datatracker.ietf.org/doc/html/rfc8200)? I'm not sure what the general MTU support available is for networking equipment but looking at my Macbook, for example, it lists 1500 as "standard" MTU for both a USB ethernet link I have and the internal wifi radio.

specification/transport/udp/udp.tex
```diff
@@ -0,0 +1,370 @@
...
+Limiting the MTU at this value allows nodes that do not transmit and/or receive transfers larger than 480~bytes
+to omit support for multi-frame transfer decomposition and/or reassembly.
+
+Nodes emitting Cyphal/UDP traffic \emph{should} avoid IP fragmentation by
+limiting the size of the UDP payload accordingly\footnote{
+    Transfers larger than the MTU limit will be transmitted as multi-frame Cyphal transfers
+    (section~\ref{sec:transport_transfer_payload}).
+    The preference towards Cyphal framing over IP framing is to remove the limitation on the maximum transfer size
+    imposed by the UDP protocol and to permit preemption of long transfers by higher-priority transfers.
+    Further, the support for reassembly of large fragmented IP packets is not universal.
```
thirtytwobits, line 367: Let's reference this RFC -> https://datatracker.ietf.org/doc/html/rfc8900 <- and just say "Cyphal does not require IP framing per the recommendations of RFC8900"

specification/transport/udp/udp.tex
```diff
@@ -0,0 +1,370 @@
...
+
+In this section, the maximum transmission unit (MTU) is defined as the maximum size of a UDP/IP datagram payload.
+
+This specification does not restrict the MTU of the underlying transport.
+It is recommended, however, to avoid MTU values less than 508~bytes,
+allowing applications to exchange up to $508 - 24 - 4 = 480$ bytes of payload in a single-frame transfer.
+Limiting the MTU at this value allows nodes that do not transmit and/or receive transfers larger than 480~bytes
+to omit support for multi-frame transfer decomposition and/or reassembly.
+
+Nodes emitting Cyphal/UDP traffic \emph{should} avoid IP fragmentation by
```
thirtytwobits, line 361: We need to be a bit stronger here since the specification must provide a reasonable amount interoperability guarantees. We should state that Cyphal/UDP end systems shall not utilize IP fragmentation on transmission and shall not be required to support it for reception. If routers end up doing it in a manner that is opaque to the end systems then the specification makes no prohibitions, however; any router that Cyphal/UDP end systems are connected to SHOULD _not_ send ICMP messages indicating a transmission is too large because:

1. This increases the network load which can modify the timing of the system and which may be considered an attack vector for some systems.
3. Cyphal/UDP nodes are not required to support ICMP messaging and will therefore not be able to modify their behaviour based on PTB messages.

specification/transport/udp/udp.tex
pavel-kirienko: Indeed, 1500 bytes is standard for Ethernet, and I am fairly confident that 1280 will work anywhere unless you want to shuttle your packet across the public Internet, but there is another concern we should be considering: the worst-case preemption latency, which is a function of the link speed and MTU. You may remember that AFDX limits the MTU per VL type for this reason. By requiring (or recommending) a higher MTU we may strip the designer of the flexibility required to meet the timing constraints. You might argue here that this is less relevant for Cyphal/UDP as there is Cyphal/TSN on the horizon, but unless there are strong reasons to do otherwise, we should not undermine the ability of the protocol to meet stricter real-time requirements.

thirtytwobits (approved):
Cool!

PR #128, opened 2023-05-11
Add Cyphal/serial

### LINKED ISSUE #99: Specify Cyphal/serial
The draft is available at https://pycyphal.readthedocs.io/en/stable/api/pycyphal.transport.serial.html

The discussion is at:

- https://forum.uavcan.org/t/uavcan-serial-issues-with-dma-friendliness-and-bandwidth-overhead/846
- https://forum.uavcan.org/t/alternative-transport-protocols-in-uavcan/324

pavel-kirienko:
Closes #99


maksimdrachov:
Trying to test the example you gave here: https://github.com/maksimdrachov/test-serial-pr

Coming up with some inconsistency, the output of running this script:

```
serial_frame:  SerialFrame(priority=NOMINAL, transfer_id=0, index=0, end_of_transfer=True, payload=303132333435363738, source_node_id=1234, destination_node_id=1, data_specifier=MessageDataSpecifier(subject_id=1234), user_data=0)
memoryview_output:  b'\x00\x06\x01\x04\xd2\x04\x01\x03\xd2\x04\x01\x01\x01\x01\x01\x01\x01\x01\x01\x01\x02\x80\x01\x0c\xbd\x88012345678\x00'
(starting delimiter) memoryview_bytes[0]:  b'\x00'
(COBS overhead byte) memoryview_bytes[1]:  b'\x06'
(header) memoryview_bytes[2:2+24]: 
b'\x01'
b'\x04'
b'\xd2'
b'\x04'
b'\x01'
b'\x03'
b'\xd2'
b'\x04'
b'\x01'
b'\x01'
b'\x01'
b'\x01'
b'\x01'
b'\x01'
b'\x01'
b'\x01'
b'\x01'
b'\x01'
b'\x02'
b'\x80'
b'\x01'
b'\x0c'
b'\xbd'
b'\x88'
(payload) memoryview_bytes[27:27+11]: 
b'1'
b'2'
b'3'
b'4'
b'5'
b'6'
b'7'
b'8'
b'\x00'
COBS-encoded transfer-CRC b''
Traceback (most recent call last):
  File "/Users/maksimdrachov/test-serial-pr/test-serial-pr.py", line 72, in <module>
    "(ending delimiter) memoryview_bytes[44]: ", memoryview_bytes[44].to_bytes(1, "big")
                                                 ~~~~~~~~~~~~~~~~^^^^
IndexError: index out of range
```

Note: I think you forgot to mention which `destination_node_id` is used?

pavel-kirienko:
There might be an off-by-one error in your snippet. You can try and cross-validate it using these instructions:

https://github.com/OpenCyphal/specification/blob/4d80115dbeb3fb9d3277932510a2df45c1758cea/specification/transport/serial/serial.tex#L158-L164

maksimdrachov:
If it's not using pycyphal I'm not interested.

maksimdrachov:
Using Yakut:

image

image

Seems to aling.

maksimdrachov:
Second example, using yakut:

image

image

Doing something wrong here?

pavel-kirienko:
xxd groups output by 16 bytes. Send it more data to see the full buffered
output.

maksimdrachov:
After re-sending the same empty message:

image

pavel-kirienko:
The pointed-to element corresponds to the source node-ID, which is 1234 = 0x04d2.

maksimdrachov:
image

maksimdrachov:
> Ah, good catch! I changed the source node-ID to 4321 but failed to reflect this in the text. Fix incoming.

image

maksimdrachov:
The ones marked in purple are also different from the text:

image

_Looks like the first 2 need to be switched around?_

pavel-kirienko:
If you could just push a commit updating the examples that would be dandy

pavel-kirienko:
Yes, Yakut uses PyCyphal under the hood.

maksimdrachov:
I think there's some issue with the current implementation of pycyphal. Resulting in a different result between yakut, which is using an older version and pycyphal.

See unit test in version 1.8:

https://github.com/OpenCyphal/pycyphal/blob/d2a7eb923bd8cbeb941e9032494ade48e755654e/pycyphal/transport/serial/_frame.py#L199-L237

And compare with unit test in the latest version:

https://github.com/OpenCyphal/pycyphal/blob/4f56e6518194268c7a6ae6c4b232e5a7aead7005/pycyphal/transport/serial/_frame.py#L230-L268

I have update the script [here](https://github.com/maksimdrachov/test-serial-pr) and even the length of the resulting frame is shorter (36 vs 42)

maksimdrachov:
image

image

maksimdrachov:
Ok, I have taken a look at how the `pub` functionality is implemented in yakut.

It calls `send()` on `SerialOutputSession` (which makes sense).

Now in `pycyphal/serial`, there seem to be 2 different implementations of `send`, which was what caused my initial confusion (see `send_tranfer()` in `SerialTransport`).

Two methods next to each other:

image


I checked in `UDPTransport` and it doesn't have any methods related to sending. Can you comment on this?

pavel-kirienko:
>Two methods next to each other:

They look okay to me.

>I checked in UDPTransport and it doesn't have any methods related to sending. Can you comment on this?

This is because in Cyphal/UDP, sending is done via sessions by writing directly to the socket.

maksimdrachov:
Ok, we can discuss it next time we meet. I'm still a bit confused.

maksimdrachov (dismissed):

specification/transport/serial/serial.tex
```diff
@@ -0,0 +1,231 @@
...
+
+\subsection{Framing}
+
+Cyphal/serial uses the ``Consistent Overhead Byte Stuffing'' (COBS) encapsulation method\footnote{%
+    See ``Consistent Overhead Byte Stuffing'', Stuart Cheshire and Mary Baker.
+    The COBS overhead is 1 byte in every 254 bytes of encapsulated data, which is about 0.4\%.
+} with zero byte as the frame delimiter.
+Due to the nature of COBS, the frame delimiter will not appear in the frame payload.
+A frame delimiter may terminate a frame and/or indicate the start of a new frame.
+The number of frame delimiters between adjacent frames may exceed one.
```
thirtytwobits, line 51: What about bit-level framing? Do we want to take advantage of the proposals in the Cheshire and Baker paper for this? If so then the header would start with `0x80, 0x00` (If I'm reading the paper correctly)

specification/transport/serial/serial.tex
```diff
@@ -0,0 +1,231 @@
...
+\end{figure}
+
+A frame consists of two parts:
+the fixed-size header (section~\ref{sec:transport_serial_header})
+immediately followed by the payload (section~\ref{sec:transport_serial_payload}).
+The header contains a dedicated header-CRC field which allows implementations to detect frame corruption early.
+
+Neither the underlying medium nor the Cyphal/serial transport layer impose any restrictions on the maximum frame size,
+which allows all Cyphal/serial transfers to be single-frame transfers\footnote{%
+    Lack of multi-frame transfers simplifies implementations significantly.
```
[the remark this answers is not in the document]
pavel-kirienko: It is placed in the footnote, not in the main text. Footnotes are used for non-normative statements throughout the text along with the blue boxes. Changing this would require updating most of the text so I am inclined to keep it.

specification/transport/serial/serial.tex
pavel-kirienko: Bit-level framing is not applicable to byte-level links such as serial port or TCP that this transport is designed for. Unless I misunderstand you.

thirtytwobits: Is that worth discussing in the specification? That Cyphal/Serial expects bit-level framing to be provided by a lower layer? (e.g. you need to use start/stop bits over RS232)

specification/transport/serial/serial.tex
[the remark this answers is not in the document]
thirtytwobits: It's still unquantified. I don't like have things like "it's great!" or "that would be bad" in our specification. Instead how about: "Omitting multi-frame transfers as a requirement is expected to simplify implementations."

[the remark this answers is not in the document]
pavel-kirienko: Updated.

specification/transport/serial/serial.tex
pavel-kirienko: The current text states that Cyphal/serial is for byte-level transports, which necessarily implies that lower-level framing is already provided:

>This section specifies a concrete transport that operates on top of raw byte-level communication channels, such as TCP/IP connections, SSL, UART, RS-232, RS-422, USB CDC ACM, and any similar communication links that allow exchange of unstructured byte streams.

thirtytwobits (approved):
image

PR #141, opened 2025-05-09
Address issues #129 #131 #139 #140

### LINKED ISSUE #129: Indicate that node-ID values 126 and 127 are reserved for diagnostic tools irrespective of the maximum node-ID value
https://forum.opencyphal.org/t/reserved-identifiers-for-diagnostic-and-debugging-tools/1833

### LINKED ISSUE #131: Cyphal/UDP: MTU shall remain constant for the duration of a transfer
Context: https://matrix.to/#/!iIoEluFjwtvXWxhOJg:matrix.org/$TM4A46iRgBFc64yi1hev5Uuam3fAsZGgvMpD-vunkfc?via=matrix.org&via=lenfesty.ca&via=stratum0.org

Also, full MTU utilization should be required for all frames of a transfer except the last one. MTU discovery is required for out-of-order reassembly.

### LINKED ISSUE #139: Random transfer-ID initialization
Tracking https://forum.opencyphal.org/t/improve-the-transfer-id-timeout/2375

### LINKED ISSUE #140: The transfer-ID timeout of RPC response RX ports should be zero
This is something that we discussed while working on libcyphal. With a non-zero transfer-ID timeout, responses may be lost if the server responds to multiple requests not in the order of their arrival. There is no risk of response duplication because the client will remove the pending request entry once its first response is received.

pavel-kirienko:
Closes #129 
Closes #131 
Closes #140

### DIFF
--- a/specification/application/conventions.tex
+++ b/specification/application/conventions.tex
@@ -12,8 +12,8 @@ An overview of related concepts is provided in chapter~\ref{sec:basic}.
 Valid values of node-ID range from 0 up to a transport-specific upper boundary
 which is guaranteed to be above 127 for any transport.
 
-The two uppermost node-ID values are reserved for diagnostic and debugging tools;
-these node-ID values should not be used in fielded systems.
+Node-ID values of 126 and 127 are reserved for diagnostic and debugging tools.
+These values should not be used in fielded systems.
 
 \subsection{Service latency}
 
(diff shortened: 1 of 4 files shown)


thirtytwobits (changes requested):
I don't think we should change the transfer ID in 1.0. Let's move this to 1.1 so we have more time to discuss. I think it's a slight improvement but there's some opportunity for refinement I'd like to explore.

pavel-kirienko:
But then should we keep the 1M count leap backward clause?

scottdarch:
> But then should we keep the 1M count leap backward clause?

No. I think we should just leave the transfer ID as is for 1.0. It's a fait accompli for anything existing anyway.

thirtytwobits (approved):

PR #142, opened 2025-05-09
Add UTF-8 and byte strings

pavel-kirienko:
Closes https://github.com/OpenCyphal/specification/issues/51

See https://github.com/OpenCyphal/pydsdl/pull/97


thirtytwobits (approved):
I approve but I think the copilot suggestion is correct, that there's a misspelling?

scottdarch (approved):

Issue #73, opened 2020-02-29
Autopublish


thirtytwobits:
Use gh-pages to automatically publish the specification after a build. See https://github.com/UAVCAN/libuavcan/blob/uavcan-v1.0/test/ci/postbuild-upload-docs.sh for an example.

pavel-kirienko:
Publish where?

thirtytwobits:
exactly!

PR #145, opened 2025-05-16
deploy specification to gh-pages

songmeo:
addresses #73

Need to rename "specification" to "docs" because gh-pages only pick up /(root) or /docs.

The doc is uploaded to http://opencyphal.github.io/specification/Cyphal_Specification.pdf.


songmeo:
@pavel-kirienko I addressed your comment. The folder is reverted back to specification. I found another option to deploy pages without renaming it. The doc is still uploaded to http://opencyphal.github.io/specification/Cyphal_Specification.pdf like before.

.github/workflows/build.yml
```diff
@@ -40,3 +40,21 @@ jobs:
         name: workspace-${{github.job}}
         path: '*'
         retention-days: 3
+
+    - name: Upload GitHub Pages artifact
```
pavel-kirienko, line 44:
```suggestion
    - name: Publish the specification PDF; the published file is linked by the main website via https://opencyphal.org/specification
```

pavel-kirienko (approved):
very good

